PR #9, opened 2017-12-04
fix(custom-decorators): Missing export word

Caballerog:
If a reader copy/paste the example of custom-decorators, it require the reserved word "export" when it want use the custom-decorator (import from another file).


kamilmysliwiec:
Right, thanks!!!

PR #10, opened 2017-12-05
fix(custom-decorators): Conflict

Caballerog:
There are a conflict between custom-decorator called User and UserEntity (interface) called User. Any reader would get an error when import User from custom-decorator and User from Entity (or interface).


kamilmysliwiec:
Hi @Caballerog,
Oh. True 👍 Thanks! 😄

PR #11, opened 2017-12-05
Update controllers.component.ts mistype

LucasBrd:
Change for HttpCode instead of HttpStatus in get statusCode()


kamilmysliwiec:
Hi @LucasBrd,
ahh. Thanks!!! 😃

PR #12, opened 2017-12-10
Pipes: (fix) deprecated import of HttpException

VinceOPS:
Hi,

I've just been through this part of the tutorial and noticed that a deprecated import was used.
Indeed, `HttpException` is now part of `@nestjs/common` instead of `@nestjs/core`.
Moreover, as a dedicated exception class now exists for this particular usecase (`BadRequestException`), I just thought it would be a good idea to use it.

### DIFF
--- a/src/app/homepage/pages/pipes/pipes.component.ts
+++ b/src/app/homepage/pages/pipes/pipes.component.ts
@@ -64,8 +64,7 @@ export class CreateCatDto {
 
   get fullValidationPipe() {
     return `
-import { HttpException } from '@nestjs/core';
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
Hi @VinceOPS,
On this page the `HttpException` is used 2 times, so if we wanna change it, we need to do that in the both locations  😄

VinceOPS:
Hi @kamilmysliwiec,

Indeed, I didn't pay attention! Please let me know if it's okay with the new commit.

kamilmysliwiec:
Now it's awesome! 🙂

PR #44, opened 2018-02-03
perf(): remove unused stylesheets

wbhob:
I have removed all of the empty stylesheets that were in the codebase, and their references. If contributors need them in the future, they can add them as necessary.


kamilmysliwiec:
thank you so much @wbhob 🙂

PR #45, opened 2018-02-05
Nest Router Module 🚀

shekohex:
add new Docs about [Nest Router Module](https://github.com/shekohex/nest-router) 🚦

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1204,6 +1204,7 @@
       "requires": {
         "anymatch": "1.3.2",
         "async-each": "1.0.1",
+        "fsevents": "1.1.3",
         "glob-parent": "2.0.0",
         "inherits": "2.0.3",
         "is-binary-path": "1.0.1",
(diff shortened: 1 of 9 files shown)


kamilmysliwiec:
Could you somehow reflect a similar language as rest of the docs? I mean a little bit more like a docs instead of package's advert. 

Additionally, it'd be better to somehow point on the most important things that this package provides - (1) possibility to versioning your API (for example /v1/ and /v2/) and (2) organize your route tree to avoid repetition (example with nested Dogs & Cats doesn't fit here)

shekohex:
Oh , I got it.
I will make sure to fit other docs Style and making some good example as you mentioned .
I will update this PR as soon as possible 😀

kamilmysliwiec:
btw, I think those 3 requirements are very, very important:
1. Allow `children` being an array
2. Allow `children` being ADDITIONALLY (not only) an array with just modules, for example: `children: [AuthModule, UsersModule]`
3. Make possible to **always** omit `module` property. Hence, following syntax will be possible:

```
path: 'v1',
children: [AuthModule, UsersModule]
```
^ as you can see above, there's no module on top level, and also `children` don't define their own paths

shekohex:
Hi Kamil , Actually I'm working on it, first of all I will make the routes tree endless nested array, so it will act like a tree, right now i'm figuring out how can i make it flat as possible, and it seems it works here 
image.
next i will edit this function to remap all paths as it should.
I will update this comment frequently to show what I achieved
- [ ]  Flat All Routes using an endless nested array  
- [ ] Allow children being an array with just modules
- [ ] Make possible to always omit module property

ambroiseRabier:
Any update here?
Also, unrelated, does routes like /writer/:uuid/paper/:uuid are supported ?

shekohex:
@ambroiseRabier
I think that PR is dead? IDK actually.

BTW, for your question check this https://github.com/shekohex/nest-router/issues/17

And for any related questions please open an issue there.

Regards.

BrunnerLivio:
Because of issue #1438 we need to use similar functionality as @shekohex nest-router package maintained by the nest org. Because of this parts of the PRs documentation may will get obsolete. So from my point of view it makes more sense to hold this PR on ice :/

shekohex:
@BrunnerLivio i see
Also i have no problem with that
Actually i have no problem with even adding the functionality of this module to the core i will help in the implementation too it would be fantastic.
And as a side note, as nest becoming popular this package is going bigger too, it definitely one of the most used 3rd party nest packages out there. Ans there is more people want to see this in the core, or at least moving nest-router package to the official nestjs organization.

kamilmysliwiec:
Let's close this PR. We'll connect on bringing this feature to the nest as soon as possible :)

halkliff:
@kamilmysliwiec Any updates on Nest's own Routing module?

jmcdo29:
[Check out the v8 PR](https://github.com/nestjs/nest/pull/6349). It includes the `RouterModule` code

PR #46, opened 2018-02-05
docs(@nestjs/common): Update the ValidationPipe documentation

fwoelffel:
This PR updates the `ValidationPipe` documentation according to https://github.com/nestjs/nest/pull/383
Let me know if you'd like me to add a more detailed example.

### DIFF
--- a/src/app/homepage/pages/pipes/pipes.component.html
+++ b/src/app/homepage/pages/pipes/pipes.component.html
@@ -89,7 +89,7 @@
     <span class="filename">create-cat.dto.ts</span>
     <pre><code class="language-typescript">{{ createCatDtoValidation }}</code></pre>
     <p>
-      Now it's time to add actual validation to our <code>ValidationPipe</code>.
+      Now it's time to finish the <code>ValidationPipe</code> class.
     </p>
     <span class="filename">validation.pipe.ts</span>
    <pre><code class="language-typescript">{{ fullValidationPipe }}</code></pre>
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
@fwoelffel I'd say, instead of building full `ValidationPipe`, I thought rather about just mentioning that this built-in is more robust and describe its possibilities (this's is actually done).

fwoelffel:
Hi @kamilmysliwiec
Sorry for the delayed response. Could you be more specific on your expectations for this documentation? I'll try to update this PR this weekend.

kamilmysliwiec:
In your PR you have explained step by step how to build enhanced `ValidationPipe`. I would say, instead of building full, powerful `ValidationPipe`, I thought rather about just mentioning that this built-in one is more robust. Hence we shouldn't change 'how to build `ValidationPipe`  section, leave this easier example in place, and create another section that describes possibilities of `ValidationPipe` which is built-in. It means that user would still see only a basic example (to reduce learning curve as much as possible) how to build a basic `ValidationPipe` but also will know the differences between this built-in one, and would know how to use it correctly and take advantage of it. 🙂

kamilmysliwiec:
thanks @fwoelffel ❤️

fwoelffel:
Sorry @kamilmysliwiec I just saw that in this PR I stated that the `ValidationPipe` transforms the payloads by default. I fixed this, should I submit a new PR?

PR #53, opened 2018-03-05
Update callback to throw UnauthorizedException

fmeynard:
This is to avoid inconsistency between not providing a token & providing a bad token


kamilmysliwiec:
Thanks @fmeynard! Merging

PR #66, opened 2018-04-23
Changed autoSchemaSync to synchronize (new way according to typeorm)

bartduisters:
The documentation as is (npm install --save typeorm mysql) will not work. autoShcemaSync has been changed to synchronize: http://typeorm.io/#/connection-options.


kamilmysliwiec:
Thanks :)

PR #70, opened 2018-04-29
Fix deprecated decorator for interceptors

d-yokoi:
Since version 5, the `@Interceptor()` is deprecated and `@Injectable()` is expected.


kamilmysliwiec:
Thanks 🙂

PR #72, opened 2018-05-04
Fix link to example of nestjs/mongoose module

ericzon:
The old link provided doesn't match with the documentation.


kamilmysliwiec:
Thanks :)

PR #73, opened 2018-05-10
fix scroller bug

dreambo8563:
when all the menu items opened, we can not see all of them, just make it scrollable


kamilmysliwiec:
Actually, I don't think that would help since we use not-native scroll library.

PR #74, opened 2018-05-10
Fixed ParseIntPipe

NAlexandrov:
```javascript
export class ParseIntPipe implements PipeTransform<string, number> {
  async transform(value: string, metadata: ArgumentMetadata): number {
    //                                                        ^ wrong type if is an async function
  }
}
```
transform is a async function. The return type of an async function must be the global ```Promise<T>``` type.

I removed async in PR. Async function is redundant in this case.


kamilmysliwiec:
Thanks @NAlexandrov 👍

PR #79, opened 2018-05-18
Fix incorrect routing to support page

motiejunas:
Was impossible to navigate to support subpage due to the missing slash.


kamilmysliwiec:
Thanks!

PR #147, opened 2018-09-02
docs(@nestjs) replace `npm run start` with `npm start`

indrek-rajamets:
I simply replaced `npm run start` with `npm start` in the documentation since `run` is redundant


kamilmysliwiec:
It is not always the case, it heavily depends on the OS. I believe we should stick with the whole command.

PR #161, opened 2018-09-06
update: supplemental guard injection method in the controller

muyu66:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[x] Other... Please describe: Supplementary explanation, the original explanation is not detailed enough, beginner may be not easy to understand
```

explanation is not detailed enough

Issue Number: N/A

## What is the new behavior?
explanation is detailed enough

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/src/app/homepage/pages/guards/guards.component.html
+++ b/src/app/homepage/pages/guards/guards.component.html
@@ -127,6 +127,14 @@
     <app-tabs #globalScopedGuardModuleT></app-tabs>
   </span>
   <pre><code class="language-typescript">{{ globalScopedGuardModule }}</code></pre>
+  <p>
+    Also, the guard in the controller can be similarly treated as a normal injectable class for dependency injection. E.g:
+  </p>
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
Don't you think that it would bring more confusedness? People might start thinking that putting a guard into `providers` array is required (while it's not true).

muyu66:
If I don't do this, I won't be able to inject any dependencies into the GUARD. "The provide: APP_GUARD" and "execution context" in the original text are not a reasonable solution.

kamilmysliwiec:
You'll be able to do this. Guard shouldn't be added to providers, unless you want to inject **them** somewhere.

muyu66:
Then, if i want to inject Service into GUARD, how to do.

kamilmysliwiec:
Simply inject, same as in controllers/services

Issue #88, opened 2018-06-09
Make headers linkable


mentos1386:
It would be nice if headers would be linkable, like the way it's done on github markdown. It would help out a lot when you want to show someone a section of documentation. Not you have to tell them to scroll down, but then you could just link them to that section.

kamilmysliwiec:
Great idea @mentos1386 👍

dianjuar:
This issue should be closed? Is already implemented

PR #193, opened 2018-11-07
Make headers linkable (#88)

lazarljubenovic:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] Other... Please describe:
```

No anchor headings.

Issue Number: #88 

## What is the new behavior?

Adding anchor headings.

```
[ ] Yes
[x] No
```

## Other information

Started work on this, but want to check in as I go so see if you agree with the direction. I'll need to do one of the following...

- refactor pages into a `PagesModule` and declare the `HeaderAnchorDirective` only there (I prefer this one)
- change all `<h3>` and `<h4>` in docs to `<h3 headerAnchor>` or similar

...in order to get rid of `#` which appear in the sidebar and table of contents now.

Another issue is that refreshing the page doesn't actually make it point to the anchor (which is the whole point, ha). I think this is an issue with `@angular/router` itself as I remember banging my head because of this on an old Angular project. So that would need to be fixed as well.

The TOC should also be refactored to just have a link instead of custom scrolling logic.

## Preview

image

image

image

### DIFF
--- a/src/app/common/directives/header-anchor.directive.ts
+++ b/src/app/common/directives/header-anchor.directive.ts
@@ -1,0 +1,36 @@
+import {AfterViewInit, Directive, ElementRef, Input, Renderer2} from '@angular/core';
+import {ActivatedRoute, Router} from '@angular/router';
+
+@Directive({
+  selector: 'h3,h4',
+})
+export class HeaderAnchorDirective implements AfterViewInit {
+
+  @Input('id') public overriddenId: string | undefined;
+
+  constructor (private elementRef: ElementRef<HTMLHeadingElement>,
+               private renderer: Renderer2,
+               private route: ActivatedRoute,
+               private router: Router) {
+  }
+
+  private getImplicitId (): string {
+    const content = this.elementRef.nativeElement.innerText;
+    return content.replace(/\s/g, '-').toLocaleLowerCase();
+  }
+
+  public ngAfterViewInit (): void {
+    const id = this.overriddenId == null ? this.getImplicitId() : this.overriddenId;
+    if (this.overriddenId == null) {
+      this.renderer.setProperty(this.elementRef.nativeElement, 'id', id);
+    }
+
+    const anchor = this.renderer.createElement('a');
+    this.renderer.setProperty(anchor, 'href', `${ this.router.url.split('#')[0] }#${ id }`);
+
+    const octothorpe = this.renderer.createText('#');
+    this.renderer.appendChild(anchor, octothorpe);
+    this.renderer.appendChild(this.elementRef.nativeElement, anchor);
+  }
+
+}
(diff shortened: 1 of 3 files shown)


kamilmysliwiec:
Nice! Thanks, I'll review it soon

src/app/common/directives/header-anchor.directive.ts
```diff
@@ -0,0 +1,36 @@
+import {AfterViewInit, Directive, ElementRef, Input, Renderer2} from '@angular/core';
+import {ActivatedRoute, Router} from '@angular/router';
+
+@Directive({
+  selector: 'h3,h4',
+})
+export class HeaderAnchorDirective implements AfterViewInit {
+
+  @Input('id') public overriddenId: string | undefined;
```
kiwikern: I would just call it `id` since it is only overridden if no id was explicitly defined.

src/app/common/directives/header-anchor.directive.ts
```diff
@@ -0,0 +1,36 @@
...
+               private router: Router) {
+  }
+
+  private getImplicitId (): string {
+    const content = this.elementRef.nativeElement.innerText;
+    return content.replace(/\s/g, '-').toLocaleLowerCase();
+  }
+
+  public ngAfterViewInit (): void {
+    const id = this.overriddenId == null ? this.getImplicitId() : this.overriddenId;
```
kiwikern: If the id is not yet set, I would update `this.overridenId` instead of creating another local variable id here.
(Especially, if you rename `overridenId` to `id`.)

src/app/common/directives/header-anchor.directive.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (this.overriddenId == null) {
+      this.renderer.setProperty(this.elementRef.nativeElement, 'id', id);
+    }
+
+    const anchor = this.renderer.createElement('a');
+    this.renderer.setProperty(anchor, 'href', `${ this.router.url.split('#')[0] }#${ id }`);
```
kiwikern: Use `this.id` here instead of `id`

src/app/common/directives/header-anchor.directive.ts
```diff
@@ -0,0 +1,36 @@
+import {AfterViewInit, Directive, ElementRef, Input, Renderer2} from '@angular/core';
+import {ActivatedRoute, Router} from '@angular/router';
+
+@Directive({
+  selector: 'h3,h4',
```
kiwikern: Is this selector specific enough? Do we need a way to exclude certain headings?

kiwikern:
> Another issue is that refreshing the page doesn't actually make it point to the anchor (which is the whole point, ha). I think this is an issue with `@angular/router` itself as I remember banging my head because of this on an old Angular project. So that would need to be fixed as well.

You have to enable anchorScrolling, see https://stackoverflow.com/a/52241001/4694994.

```
RouterModule.forRoot(routes, {
    anchorScrolling: 'enabled',
})
```

kamilmysliwiec:
Thanks for the inspiration. I have reused a big part of your code + scroll is fixed.

PR #213, opened 2018-12-09
useGlobalGuards clarification for websockets and microservices

dilame:
## PR Checklist

## PR Type
I think we need to notify users who reads `microservices/guards` and `websockets/guards` that `useGlobalGuards()` method doesn't set up guards for microservices and websockets, because `There is no difference` string at the beginning of the text makes user think that `useGlobalGuards()` will also work.

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[x] Other... Please describe: docs
```

```
[ ] Yes
[x] No
```

## Other information


kamilmysliwiec:
Thanks for your contribution! However, every page has been moved to the markdown syntax now (`content` directory) which makes this PR outdated. Could you please update it? (also, please notice that `useGlobalGuards()` works well for microservices - it doesn't work for hybrid apps [(n)microservices + HTTP])

PR #306, opened 2019-03-21
content: controller page edits

johnbiundo:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[x] Other... Please describe:

Doc edits for Controllers page.
```
@kamilmysliwiec 

Here's the Controllers chapter for your review.

Would you prefer getting these one at a time, or in a larger bundle?

Also, can you answer this question please?  For the Full resource sample code, shouldn't the JS version use @Bind() for the method parameter decorators?  If you want, I'll add them in.  I may have noticed a few other cases of the JS code not having @Bind() -- I'll go back and check once you confirm the correct usage.

Finally, I wonder if the last paragraph in this chapter is misleading.  It says `The first approach should always be preferred, but to make Nest backwards compatible with previous versions, the above approach is still available`.

This makes the library-specific approach sound like it could be deprecated in the future.  Since it seems that some things can only (or are best) performed with the library specific approach, I'm wondering if it's your intent to so strongly move people away from it? If not, we might consider adjusting that wording slightly.

Issue Number: N/A

```
[ ] Yes
[ ] No
```

## Other information


johnbiundo:
@BrunnerLivio 

I've forked and cloned from master into a new repo for this.

Two things I've noticed:
1) the angular app fails to compile (running `npm run start`).  It fails with a bunch of errors like this:
image

This seems maybe related to removing the HTML files from the master repo? 

It didn't really hold me back except I couldn't do the full live preview I would have liked.  What are your thoughts?

2) I notice that the Github hook checks failed on this PR unlike last time.  Could that also be related?

Thanks!

johnbiundo:
@kamilmysliwiec 

I'm wondering what you think of looking into using codesandbox for live samples, and linking to it from the docs?  I think the way Angular does it is pretty useful, and could make a nice improvement to Nest docs.  

I thought to mention it here as it might fit well into my workflow to help out with this as I continue to review/edit chapters.

BrunnerLivio:
@johnbiundo
Awesome work once again! 😁

1. You can use `npm run markdown:watch` to compile the markdown files. I think thats the error.

 I guess we should update the start script to something like this:
`concurrently "ng serve --source-map=false" "npm run markdown:watch"`
and install `npm i -g concurrently`.

2. Not quite sure: I’m on mobile - can’t see the error somehow. Maybe its a merge conflict and the upstream is not up to date?


By the way, would you mind using our git commit message standard? 
https://github.com/nestjs/docs.nestjs.com/blob/master/CONTRIBUTING.md#-commit-message-guidelines

It would be easier for us to track down past decisions if the commit messages are consistent and descriptive. 

You can find some examples in the git commit history `git log`.

johnbiundo:
@BrunnerLivio Thanks! :)

re: commit messages - sure!  I'll do that going forward.

re: angular app errors - tried your suggestion.  To clarify, I think the markdown compiler is working fine.  It's the live preview that isn't working.  I get the same compile errors when running it with concurrently :(

kamilmysliwiec:
```
7:35:22 PM: > ng build --prod --aot --deleteOutputPath=false
7:35:35 PM: Date: 2019-03-21T18:35:35.533Z
7:35:35 PM: Hash: 7ad75200d033954ed2e8
7:35:35 PM: Time: 9389ms
7:35:35 PM: chunk {0} runtime.26209474bfa8dc87a77c.js (runtime) 1.41 kB [entry] [rendered]
7:35:35 PM: chunk {1} main.01ecbc0fcff9207e8cde.js (main) 128 bytes [initial] [rendered]
7:35:35 PM: chunk {2} polyfills.a723c36635c3a098c2aa.js (polyfills) 130 bytes [initial] [rendered]
7:35:35 PM: chunk {3} styles.b2a8a06db2fdc42df061.css (styles) 73.1 kB [initial] [rendered]
7:35:35 PM: chunk {scripts} scripts.c4e6a0cd08523fd1390d.js (scripts) 17.3 kB [entry] [rendered]
7:35:35 PM: ERROR in : Template parse errors:
7:35:35 PM: Unexpected character "EOF" (Do you have an unescaped "{" in your template? Use "{{ '{' }}") to escape it.) ("ing us to have full control of the response object (headers manipulation and so on).</p>
7:35:37 PM: failed during stage 'building site': Build script returned non-zero exit code: 1
7:35:35 PM:     </div>
7:35:35 PM: [ERROR ->]"): /opt/build/repo/src/app/homepage/pages/controllers/controllers.component.html@433:0
```

it seems that there is an angular build error.

johnbiundo:
OK, found it.  I had an unescaped { in a code snippet I added.  I'll push the fix shortly.

@BrunnerLivio: I think this also fixed my live preview problem!

kamilmysliwiec:
Thanks for another amazing contribution @johnbiundo 🎉 

> Also, can you answer this question please? For the Full resource sample code, shouldn't the JS version use @Bind() for the method parameter decorators? If you want, I'll add them in. I may have noticed a few other cases of the JS code not having @Bind() -- I'll go back and check once you confirm the correct usage.

You are right, my bad. I had to miss that for some reason.

> This makes the library-specific approach sound like it could be deprecated in the future. Since it seems that some things can only (or are best) performed with the library specific approach, I'm wondering if it's your intent to so strongly move people away from it? If not, we might consider adjusting that wording slightly.

It won't be deprecated. Nonetheless, it shouldn't be preferred due to its incompatibility with interceptors (you cannot intercept returned values then), `@HttpCode()` decorator, and is not platform-agnostic (different HTTP libraries may vary by methods used to respond to clients). Also, it's harder to unit test (you have to mock `res` object etc).

johnbiundo:
@kamilmysliwiec Great, thanks!  I'll make that change to the sample code, and try to reflect that information into the last paragraph.

johnbiundo:
@BrunnerLivio just a note that I'm still having some intermittent issues with running the live preview of the docs.   It's not a show stopper, as I just run a markdown previewer in VS Code, which works pretty well.  It just doesn't give an actual live preview.  So I have to wait for the deploy-preview task on Github to see an actual live preview.  Doesn't really bother me, but it is possible that things may only show up in live preview (for example, the JS/TS code switcher requires a fully rendered page to test).  As more people contribute to docs, this might be more important.

If you think it's worth pursuing, I'll log a reproducible issue.

kamilmysliwiec:
Hey @johnbiundo. Simply run `npm run markdown` to generate all HTML files locally - it will fix the build.

johnbiundo:
@kamilmysliwiec 👍 That did it.  Thanks!

kamilmysliwiec:
A m a z i n g changes :) Thank you!!

PR #333, opened 2019-04-01
docs(modules.md) - Edits to Modules chapter

johnbiundo:
## PR Checklist

## PR Type

- [x] Documentation

## Other information

@kamilmysliwiec Modules chapter proposed edits, for your review.

### DIFF
--- a/content/modules.md
+++ b/content/modules.md
@@ -4,7 +4,7 @@ A module is a class annotated with a `@Module()` decorator. The `@Module()` deco
 
 <figure><img src="/assets/Modules_1.png" /></figure>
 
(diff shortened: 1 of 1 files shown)


johnbiundo:
@kamilmysliwiec 

Quick meta note. Sorry for the slow down last week.

As I work on the advanced topics, I plan to do more research and testing in order to make better contributions. For example, I want to test each code sample to make sure it executes properly.  This is more time consuming (and may generate more questions to you).

After some thought, I decided to keep things moving along for now. So I'll defer that deeper testing/research and continue in "light editing" mode for the time being. As a result I may come back for a second round of editing on some chapters when I've had time to do the additional research. This is not ideal as it may require more of your review time, but I think it's best for now to keep things moving. Welcome any feedback you have on the process.

kamilmysliwiec:
Amazing contribution again, thank you!
One thing that we should clarify:
```
 In small applications, the root module may be the only module in your application.
```
We should rather emphasize that even though it's theoretically possible, it's still not considered to be a good practice.

johnbiundo:
> We should rather emphasize that even though it's theoretically possible, it's still not considered to be a good practice.

I'm on it!

kamilmysliwiec:
Amazing  🔥🔥🔥

Issue #350, opened 2019-04-09
Docs: improve sample code for global catch all exceptions


johnbiundo:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[ ] Feature request
[x] Documentation issue or request (new chapter/page)
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

Exceptions chapter, possible example code improvement.

## Expected behavior

If you were to implement the "Catch Everything" code sample in the Exception filters chapter, it would work as described.

## Minimal reproduction of the problem with instructions

Simply implement this http-exception filter (with empty `@Catch()` decorator) to handle all exceptions as shown in the docs.  Then issue a `throw new Error('boom');` in the CatsController to simulate a non-HTTP exception arising, and make the http request.  

## What is the motivation / use case for changing the behavior?

The problem is that the code sample includes the code:
`const status = exception.getStatus();`
but a vanilla `Error` class doesn't have this method.

I would propose something like the following.  If acceptable, I'll include it in an upcoming PR for this chapter.

```
import {
  ExceptionFilter,
  Catch,
  ArgumentsHost,
  HttpException,
} from '@nestjs/common';
import { Request, Response } from 'express';

@Catch()
export class HttpExceptionFilter implements ExceptionFilter {
  catch(exception: unknown, host: ArgumentsHost) {
    const ctx = host.switchToHttp();
    const response = ctx.getResponse<Response>();
    const request = ctx.getRequest<Request>();

    // default HTTP response code for non HttpException errors is
    // 500 - Internal Server Error
    let status = 500;
    if (exception instanceof HttpException) {
      status = exception.getStatus();
    }

    response.status(status).json({
      statusCode: status,
      timestamp: new Date().toISOString(),
      path: request.url,
    });
  }
}
```
## Environment

<pre><code>
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

kamilmysliwiec:
You are right, my bad. I like the idea, feel free to add this as a part of the PR.

lazarljubenovic:
It might be clearer to write

```ts
const status = exception instanceof HttpException
  ? exception.getStatus()
  : 500
```

instead of overwriting `status`, since it's only a single condition.

kamilmysliwiec:
Good point @lazarljubenovic.
[...]

PR #351, opened 2019-04-09
docs: Exception Filters doc update

johnbiundo:
## PR Checklist

## PR Type

- [x] Documentation

## Other information

@kamilmysliwiec exception-filters chapter proposed edits, for your review.

This update should be mostly independent of the following questions.  However, please look at the first question immediately below:

1) I may have made a mistake that I need a quick answer from you on:

I thought that in some cases it provided more clarity by adding @@filename() to the code example.  However, it seems that this also triggers the TS/JS switch.  I only realized this after adding that annotation to several examples, so I wanted to clarify this with you. If, by using @@filename(), we are also required to provide the @@switch directive and a JS version of the file, I'll remove that annotation.  Please let me know.

2) Following are some questions that, if you have time to answer, I will then update the PR as appropriate.  If you want to defer this to another time, I think a normal quick proofreading by you will suffice, and this PR can be processed stand-alone from them.

a) In section 'Binding filters', should we mention that the useGlobalFilters() method can take multiple filter instances?

b) In a related question, if you use the DI method (adding a provider with APP_FILTER token to a module), is there a mechanism to add multiple filters?  Assuming not, should we document that so users can consider trade-offs of the several different ways of setting up global filters?

c) In the DI method, we say 
> In order to solve this issue, you can set up a filter directly from any module using following construction

Does the 'from any module' statement imply that after setting up the filter this way, it is global across *all* modules?  It seems so, but this part of the documentation is somewhat complex, so I'd suggest we state that explicitly.

d) In the 'Inheritance' section, we say 
>  you would like to reuse an already implemented, core exception filter

What precisely do we mean by 'core exception filter'?  More specifically, is this the correct recipe for extending/overriding both BaseExceptionFilter and/or a user-defined hierarchy of filters?

e) in #350 I ask about an improvement to one of the code examples.  Let me know if you would like me to add that.


kamilmysliwiec:
> I thought that in some cases it provided more clarity by adding @@filename() to the code example. However, it seems that this also triggers the TS/JS switch. I only realized this after adding that annotation to several examples, so I wanted to clarify this with you. If, by using @@filename(), we are also required to provide the @@switch directive and a JS version of the file, I'll remove that annotation. Please let me know.

Yes, unfortunately.

> a) In section 'Binding filters', should we mention that the useGlobalFilters() method can take multiple filter instances?

This would nice!

> b) In a related question, if you use the DI method (adding a provider with APP_FILTER token to a module), is there a mechanism to add multiple filters? Assuming not, should we document that so users can consider trade-offs of the several different ways of setting up global filters?

You can add a provider with `APP_FILTER` multiple times.

> Does the 'from any module' statement imply that after setting up the filter this way, it is global across all modules? It seems so, but this part of the documentation is somewhat complex, so I'd suggest we state that explicitly.

Yes, global across all modules, BUT you can use dependencies available exclusively in the module that the filter was defined in.

> What precisely do we mean by 'core exception filter'? More specifically, is this the correct recipe for extending/overriding both BaseExceptionFilter and/or a user-defined hierarchy of filters?

Every app has a default exception filter that transforms every unrecognized exception to 500 Internal Server Error (and recognized ones - those that extend `HttpException` to appropriate objects). This default one = core exception filter.

> e) in #350 I ask about an improvement to one of the code examples. Let me know if you would like me to add that.

I'll check this out!

johnbiundo:
Got it.  OK, I'll try to amend the PR with this feedback today.

> Yes, unfortunately.

OK, That's too bad.  :(  I think it really helps readers to see where code examples should go and how to structure their files.  I'll undo those changes.

johnbiundo:
@kamilmysliwiec 

> You can add a provider with `APP_FILTER` multiple times.

I experimented with this a bit. What I seem to observe is that if you add multiple providers, the type listed in their `@Catch()` decorator should be mutually exclusive.

If I tried one `@Catch()` with a superclass (`HttpException`) and one with a subclass (`NotFoundException`), it appears that the last one registered applies, and the other is ignored.  If the classes are mutually exclusive (e.g., one for `NotFoundException` and one for `BadRequestException`), the correct one is applied.

I had originally thought the superclass/subclass approach would allow you to build a hierarchy with overrides, but that doesn't seem to be the case.

Am I understanding correctly, and do you think this should be documented?  If so, any other context you think is important to include?

kamilmysliwiec:
hey @johnbiundo, for some reason your PR affected almost every file (276 files). Could you remove these changes and include only `.md` file? :)

kamilmysliwiec:
> If I tried one @Catch() with a superclass (HttpException) and one with a subclass (NotFoundException), it appears that the last one registered applies, and the other is ignored. If the classes are mutually exclusive (e.g., one for NotFoundException and one for BadRequestException), the correct one is applied.

Yes, ordering matters in this case. We don't have any mechanism that would create a hierarchy from registered exception filters based on exceptions that they handle (btw it would very likely be even more confusing for devs I guess).

johnbiundo:
> hey @johnbiundo, for some reason your PR affected almost every file (276 files). Could you remove these changes and include only `.md` file? :)

Oops, sorry about that.  I had a little hiccup with github yesterday that resulted in me changing a global git setting.  I bet this was a side effect.  I'll fix shortly.

johnbiundo:
> Yes, ordering matters in this case. We don't have any mechanism that would create a hierarchy from registered exception filters based on exceptions that they handle (btw it would very likely be even more confusing for devs I guess).

OK, understood.  I think I'll leave it at stating that you can add multiple `APP_FILTER`s  for now.  I think it's probably worth elaborating a little, but I want to experiment more before proposing any further details.

johnbiundo:
> hey @johnbiundo, for some reason your PR affected almost every file (276 files). Could you remove these changes and include only `.md` file? :)

@kamilmysliwiec 
I can't seem to easily fix this.  I'm going to submit a new PR with these changes, then close this one.

PR #384, opened 2019-04-25
docs(custom-decorators.md) Edits for clarity and grammar

johnbiundo:
## PR Checklist

## PR Type

- [x] Documentation

## Other information

@kamilmysliwiec Custom decorators chapter proposed edits, for your review.

Most edits are minor, however I made a change to the content and flow of the 'Passing data' section that you should take a closer look at.  I thought this might flow better and demonstrate the use of the passed data more clearly, but please let me know if you agree.  If not, I'll revert those changes.

### DIFF
--- a/content/custom-decorators.md
+++ b/content/custom-decorators.md
@@ -86,40 +86,52 @@ async findOne(user) {
 
 #### Passing data
 
-When the behavior of your decorator depends on some conditions, you can use the `data` parameter to pass an argument to the decorator's factory function:
+When the behavior of your decorator depends on some conditions, you can use the `data` parameter to pass an argument to the decorator's factory function. One use case for this is a custom decorator that extracts properties from the request object by key. Let's assume, for example, that our <a href="techniques/authentication#user-object">authentication layer</a> validates requests and attaches a user entity to the request object. The user entity for an authenticated request might look like:
+
+```json
+{
+  "id": 101,
+  "firstName": "Alan",
+  "lastName": "Turing",
+  "email": "alan@email.com",
+  "roles": ["admin"]
+}
+```
+
+Let's define a decorator that takes a property name as key, and returns the associated value if it exists (or undefined if it doesn't exist, or if the `user` object has not been created).
 
 ```typescript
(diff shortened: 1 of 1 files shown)


content/custom-decorators.md
```diff
@@ -86,44 +86,44 @@ async findOne(user) {
...
-Will make possible to access the `test` string via the `data` argument:
+When the behavior of your decorator depends on some conditions, you can use the `data` parameter to pass an argument to the decorator's factory function:
 
 ```typescript
 @@filename(user.decorator)
 import { createParamDecorator } from '@nestjs/common';
 
 export const User = createParamDecorator((data: string, req) => {
-  console.log(data); // test
+  req.user.userType = data;
```
kamilmysliwiec: Custom decorators should be used to conditionally get the data rather than modify it on the fly. For instance:
```typescript
return data ? req.user && req.user[data] : req.user;
```
Then, someone can easily pick a single property of the user object using `@User('id')`

johnbiundo: Got it, that makes sense.  If I'm understanding correctly, after testing it out, this was the code sample that worked for me (slight change to yours):

```return data && req.user && req.user[data] ? req.user[data] : req.user;```

Does that make sense?  I'm returning the property based on the key passed to the decorator if it exists, otherwise the regular req.user, right?

This ability to pick a property via a decorator with passed key seems like a major point to get across (it eluded me until this discussion),  so if I'm on the right track, once this code sample is good, I'll make sure this point is clear.

Comment (kamilmysliwiec) [reply content/custom-decorators.md]: > Does that make sense? I'm returning the property based on the key passed to the decorator if it exists, otherwise the regular req.user, right?

It seems to introduce side-effects. I think that this:

```typescript
return data ? req.user && req.user[data] : req.user;
```
should be enough (just return `undefined` if the property is `undefined` 😄 )

johnbiundo:
OK, I think I understand a little better now.  Take one more look at the 'Passing data' section to see if you agree with this change.

kamilmysliwiec:
Great example! Thank you ❤️

PR #433, opened 2019-05-22
feat(A11Y): improve a11y on toc component

dianjuar:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] Other... Please describe:
```

Currently the toc component has two problems related with a11y (accessibility):
- Is not focusable
- Does not indicate where you are going to go (you can not the preview URL)

## What is the new behavior?
The new improvements are:
- The toc elements are focusable
- You can see where you are going to go (you can see the URL)
- When you select a title, the focus is transferred to the title to continue browsing
- The focus flow is changed
  - image
  - Now you pass from the navbar (N° 2) to the toc component (N° 3), because has no sense to pass through all the article to go there.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/styles.scss
+++ b/src/styles.scss
@@ -4,6 +4,10 @@
 @import './scss/variables.scss';
 @import './scss/utils.scss';
 
+:root {
+  scroll-behavior: smooth;
+}
+
 body {
   font-size: 16px;
   font-weight: 400;
--- a/src/app/homepage/homepage.component.html
+++ b/src/app/homepage/homepage.component.html
@@ -12,8 +12,8 @@
   <app-menu [class.opened]="isSidebarOpened" perfectScrollbar></app-menu>
   <div class="container" [class.wide]="!isSidebarOpened">
     <div class="page-wrapper">
-      <router-outlet (activate)="onRouteActivate($event)"></router-outlet>
       <app-toc [contentReference]="contentRef"></app-toc>
+      <router-outlet (activate)="onRouteActivate($event)"></router-outlet>
     </div>
     <div class="sponsors-wrapper">
       <div class="sponsors-container">
--- a/src/app/shared/components/toc/toc.component.html
+++ b/src/app/shared/components/toc/toc.component.html
@@ -4,9 +4,10 @@
       *ngFor="let item of headings; trackBy: trackById; let index = index"
       [attr.id]="item.id"
       [class.current]="index === activeId"
-      (click)="navigateToAnchor(item.elementRef)"
-    >
-      {{ item.textContent }}
+      >
+      <a href="{{ currentURL() }}#{{ item.elementRef.id }}">
+        {{ item.textContent }}
+      </a>
     </li>
   </ul>
 </div>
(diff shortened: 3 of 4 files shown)


src/styles.scss
```diff
@@ -4,6 +4,10 @@
 @import './scss/variables.scss';
 @import './scss/utils.scss';
 
+:root {
+  scroll-behavior: smooth;
```
kamilmysliwiec: Isn't it too early to use this property?

dianjuar: Because is not fully [implemented](https://caniuse.com/#search=scroll-behavior) ?

dianjuar: Quoting the css tricks article [_Smooth Scrolling and Accessibility_](https://css-tricks.com/smooth-scrolling-accessibility/#article-header-id-0)

>  when you follow a link, the keyboard focus should follow it, too and be able to access the next element after the target. 

Here is an example

### [demo](https://output.jsbin.com/yilokec)
example

#### To implement a smooth scrolling with that recommendation the best way that I could find is with that css property. 

----------------

## Why?
The current implementation of the smooth scrolling do not update the keyboard focus.
```js
elementRef.scrollIntoView({	
  behavior: 'smooth',	
  block: 'start'	
});
```

There is a lack of callback to know when the scroll finish to update the focus to the new element and if you do it right away like this
```js
elementRef.scrollIntoView({	
  behavior: 'smooth',	
  block: 'start'	
});
elementRef.focus();
```
When you `elementRef.focus()`, the target hijacks the smooth scrolling and it jumps right at the target, abruptly.

----------------

The possible solutions that I found were:
- Use Jquery animate
- Implement a `setTimeout` with the same duration as the scrolling
- Use the css property `:root { scroll-behavior: smooth; }`

I don't want to add a library like jquery just for that, making the web page load slow.
Agreeing with a [comment](https://css-tricks.com/smooth-scrolling-accessibility/#comment-1608095) in the article, use a `setTimeout` doesn't feel particularly elegant.

If you look at [Can I Use](https://caniuse.com/#search=scroll-behavior) the porcentaje of usage is 71.42% considering all users (desktop and mobile), doesn't look too bad like the other options.

image

dianjuar: There is something that is not clear o I need to go deeper ?

kamilmysliwiec: The idea is great. I'm just afraid about browser compatibility for now. 71% isn't that much

dianjuar: The 29% left, just is not going to have a smooth scroll, the functionality is going to be the same. Their jump is going to be immediate

dianjuar: You can make the test in an old browser (chrome 59) that doesn't support the css property

https://www.browserling.com/browse/win/7/chrome/59/https%3A%2F%2Fdeploy-preview-433--docs-nestjs.netlify.com%2F

dianjuar: It rises up to [75.97%](https://caniuse.com/#search=scroll-behavior)

kamilmysliwiec:
Thanks!

Issue #456, opened 2019-06-04
Decorators for req and res


NickBolles:
## I'm submitting a...
<pre><code>
[ ] Regression
[ ] Bug report
[ ] Feature request
[c] Documentation issue or request (new chapter/page)
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior
On the [controllers page](https://docs.nestjs.com/controllers) I think the table for the provided decorators and the objects they represent is wrong for Request and Response. 

## Expected behavior
The should be @Req and @Res I believe, not @Request @Response. That's what is used in the examples

kamilmysliwiec:
Actually, both are valid (`@Res()` and `@Response`). We should probably mention that

kamilmysliwiec:
Done

PR #535, opened 2019-07-22
docs(controllers.md) Add note on @Res() per #456

johnbiundo:
## PR Checklist

## PR Type

- [x] Documentation

## Other information

@kamilmysliwiec Controller page update re: issue #456 for your review.

### DIFF
--- a/content/controllers.md
+++ b/content/controllers.md
@@ -103,11 +103,10 @@ The request object represents the HTTP request and has properties for the reques
   <tbody>
     <tr>
       <td><code>@Request()</code></td>
-      <td><code>req</code></td>
-    </tr>
+      <td><code>req</code></td></tr>
     <tr>
       <td><code>@Response()</code></td>
-      <td><code>res</code></td>
+      <td><code>res</code><span class="table-code-asterisk">*</span></td>
     </tr>
     <tr>
       <td><code>@Next()</code></td>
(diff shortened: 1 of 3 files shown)


johnbiundo:
Fixed bad HTML

kamilmysliwiec:
Should we add some CSS to make this * more visible?

image

johnbiundo:
Agreed!  How about this (proposed new on the right)?
image

Here's a bigger view:
image

Above also has removed `<sup> </sup>` tags, so the * is lower.  Here it is **with** `<sup>` back in:

image

Here's the CSS:
`style="color: #2876d2; font-weight: 700;"`


RE: CSS - I could do it inline, or add a class to `styles.css`.  Or leave it to someone else if they own Docs styles.

johnbiundo:
I think `<sup>` overrode font-weight on last sample.  We could keep the larger size AND just do vertical-align (what `<sup>` does anyway) to get both.

kamilmysliwiec:
Imo this one is the best:

image

(without sup)

> CSS - I could do it inline, or add a class to styles.css.

Feel free to add a class:)

johnbiundo:
Sorry about the multiple commits.  I haven't quite got the hang of squash yet.

kamilmysliwiec:
Np. I can see some conflicting files. Could you resolve them, please? :) @johnbiundo

johnbiundo:
> Np. I can see some conflicting files. Could you resolve them, please? :) @johnbiundo

@kamilmysliwiec should be good now.

kamilmysliwiec:
thank you :)

PR #596, opened 2019-08-20
bugfix(docs) adjust toc height

johnbiundo:
## PR Checklist

## PR Type

- [x] Documentation

## Other information

@kamilmysliwiec Addresses #587 Sets max height for TOC section to 600px;

### DIFF
--- a/src/app/shared/components/toc/toc.component.scss
+++ b/src/app/shared/components/toc/toc.component.scss
@@ -8,7 +8,7 @@
   position: fixed;
   top: 90px;
   right: 20px;
-  max-height: 350px;
+  max-height: 600px;
 
   @media print {
     display: none;


src/app/shared/components/toc/toc.component.scss
```diff
@@ -8,7 +8,7 @@
   position: fixed;
   top: 90px;
   right: 20px;
-  max-height: 350px;
+  max-height: 600px;
```
BrunnerLivio:
```suggestion
    max-height: 100%;
    min-height: 350px;
```

Wouldn't that make more sense? :)

kamilmysliwiec:
It's actually hard to say if 600px is OK. This value should be calculated based on the window size - ad height

Issue #615, opened 2019-08-27
Kafka Microservice Documentation


mkaufmaner:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[ ] Feature request
[x] Documentation issue or request (new chapter/page)
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## What is the motivation / use case for changing the behavior?

Documentation detailing how to implement the Kafka microservice.

See https://github.com/nestjs/nest/issues/2361
See https://github.com/nestjs/nest/pull/2735
See https://github.com/nestjs/nest/pull/2735#issuecomment-525151470

## Environment

<pre><code>
For Tooling issues:
- Node version: N/A  
- Platform:  N/A 

Others: N/A

</code></pre>

ArkeologeN:
@mkaufmaner - it seems like the documentation ticket is long pending. It would be nicer if we could publish it, so finding the implementation examples would be easier :- )

jaligamap:
@mkaufmaner - is this feature implemented? If yes, could you please provide some examples on how to use it?

rubiin:
i would very much like to see docs on kafka like other microservices

petarpredragperica:
Eagerly awaiting the docs, for now the idea behind Kafka implementation is not quite understandable from the existing docs.

jakagacic:
The consuming of messages does not seem to be working unless I pass the same config to both the server and the client. What is the role of the kafka server created by `NestFactory.createMicroservice()`?

jakagacic:
I think I figured this one myself - the server side is used for consuming messages, and the client side for sending message (events). Am I right? If this is the case, why does the client also create an instance of a consumer?

XBeg9:
any updates on documentation? Thanks :)

PR #618, opened 2019-08-27
feat(microservice) Kafka Documentation

mkaufmaner:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] Other... Please describe:
```

Issue Number: #615 

```
[ ] Yes
[x] No
```

## Other information


kamilmysliwiec:
@mkaufmaner we should probably add a note about `subscribeToResponseOf()` usage (it's not required by any other transport strategy). Additionally, it's probably worth mentioning that `ClientKafka` should be used, instead of generic `ClientProxy` (similar case as in `gRPC`). Lastly, it would be a valuable addition if we provided some info about the capability to pass `key` to `send()` method

mkaufmaner:
> @mkaufmaner we should probably add a note about `subscribeToResponseOf()` usage (it's not required by any other transport strategy). Additionally, it's probably worth mentioning that `ClientKafka` should be used, instead of generic `ClientProxy` (similar case as in `gRPC`). Lastly, it would be a valuable addition if we provided some info about the capability to pass `key` to `send()` method

@kamilmysliwiec I will certainly update the documentation and this is still a work in progress.  Have been busy diving into analytics engines the past week so haven't had much time to work on it.

kamilmysliwiec:
Sure thing @mkaufmaner. Thank you!

mkaufmaner:
Many thanks to @johnbiundo for his changes!  This should be ready to go now.

content/microservices/kafka.md
```diff
@@ -0,0 +1,283 @@
...
+```
+
+#### Outgoing
+
+Nest sends outgoing Kafka messages after a serialization process when publishing events or sending messages. This occurs on arguments passed to the client `emit()` and `send()` methods or on values returned from a `@MessagePattern` method. This serialization "stringifies" objects that are not strings or buffers by using `JSON.stringify()` or the `toString()` prototype method.
+
+```typescript
+@@filename(hero.controller)
+@Controller()
+export class HeroService {
```
alfredoperez: Was this supposed to be `HeroService` or `HeroController`

content/microservices/kafka.md
```diff
@@ -0,0 +1,283 @@
...
+    ];
+
+    return {
+      headers: {
+        realm
+      },
+      key: heroId,
+      value: items
+    }
+  }
```
alfredoperez: I saw an example of outgoing messages on the tests that use the  `client` `emit()` and `send()`. Is this another way to send messages. Which one is preferred?

content/microservices/kafka.md
```diff
@@ -0,0 +1,283 @@
...
+      groupId: 'hero-consumer' // hero-consumer-client
+    }
+  }
+})
+client: ClientKafka;
+```
+
+> info **Hint** Kafka client and consumer naming conventions can be customized by extending `ClientKafka` and `KafkaServer` in your own custom provider and overriding the constructor.
+
+Since the Kafka microservice message pattern utilizes two topics for the request and reply channels, a reply pattern should be derived from the request topic. By default, the name of the reply topic is the composite of the request topic name with `.reply` appended.
```
alfredoperez: I have not been able to listen to messages/event when only using this client.

Do I need to always create the microservice  (NestFactory.createMicroservice()) to listen to messages?

Can I have different consumers per application or creating the microservice will make it to always use the `Kafka Server`?

alfredoperez: I think my problem was that I was trying to  listen to events from service and not a controller

alfredoperez:
@mkaufmaner is it possible to have multiple controllers responding to the same incoming message or event?

MikeyUchiha:
@mkaufmaner what's the status on this? i would love to be able to use this but no documentation.

content/microservices/kafka.md
jakagacic: @kamilmysliwiec I have a similar question - the consuming of messages does not seem to  be working unless I pass the same config  to both the server and the client. What is the role of the kafka server created by NestFactory.createMicroservice()?

PR #701, opened 2019-09-25
docs(CLI) New CLI chapter

johnbiundo:
## PR Checklist

## PR Type

- [x] Documentation

## Other information

@kamilmysliwiec This is a draft for early feedback from you.  The `usages` page will probably be repurposed into a command reference with options and description of each command.  Also envision another page for *libraries* with details about how libraries should be designed and how they are compiled and tsconfig paths are resolved.

Anyway, would love your feedback before going much further.


content/cli/overview.md
```diff
@@ -1,23 +1,79 @@
 ### Overview
 
-[Nest CLI](https://github.com/nestjs/nest-cli) is a command-line interface tool that helps you to initialize and develop your applications. It helps on many grounds at once, from scaffolding the project to building well-structured apps. The Nest CLI is based on the [Angular Devkit](https://github.com/angular/devkit) package.
+The [Nest CLI](https://github.com/nestjs/nest-cli) is a command-line interface tool that helps you to initialize, develop, and maintain your Nest applications. It assists in multiple ways, including scaffolding the project, serving it in development mode, and building and bundling the application for production distribution.  It embodies best-practice architectural patterns to encourage well-structured apps.
+
+The Nest CLI is based on the [Angular Devkit](https://github.com/angular/devkit) package, and works for server-side applications in much the same way that the [Angular CLI](https://angular.io/cli) assists development of client-side Angular apps.
```
kamilmysliwiec: > The Nest CLI is based on the [Angular Devkit](https://github.com/angular/devkit) package

I think that we should remove this sentence. Basically, only `generate` commands are using Angular Devkit underneath, nothing else

content/cli/overview.md
```diff
@@ -1,23 +1,79 @@
...
+
+#### Workspaces
+
+When you run `nest new`, Nest generates a boilerplate application structure by creating a new folder and populating an initial set of files.  You can continue working in this default structure, adding new components, as described throughout this documentation.  We refer to this as **single project mode**.  Nest also supports the notion of a multi-project **workspace mode**.  You can read the section on [Workspaces](/cli/workspaces) to decide which mode is most suitable for you.  Aside from a very few specific considerations around how the **build** process works (essentially, workspaces automate the build complexities that can sometimes arise from monorepo-style project structures), the rest of this documentation applies equally to both single project and workspace mode approaches.  In fact, you can easily switch from single project mode to workspace mode at any time in the future, so you can safely defer this decision while you're still learning about Nest.
```
kamilmysliwiec: We should add a note here that if you want to simply develop multiple applications that would never share anything/separate libraries that will be eventually published to NPM - multiple workspaces can be used (separate dependencies `node_modules`, separate package.json etc).

content/cli/workspaces.md
```diff
@@ -0,0 +1,206 @@
+### Workspaces
+
+Nest has two modes for organizing code:
+- **single project mode**: useful for building individual project-focused applications that have their own dependencies and don't need to optimize for sharing modules, or optimizing complex builds.  This is the default mode.
```
kamilmysliwiec: Not sure about the naming here (**single project mode**), but maybe I'm overreacting :D

content/cli/workspaces.md
```diff
@@ -0,0 +1,206 @@
...
+
+The `nest-cli.json` file looks like this:
+```json
+{
+  "collection": "@nestjs/schematics",
+  "sourceRoot": "src"
+}
+```
+
+We can convert this to a *workspace mode* structure as follows:
```
kamilmysliwiec: We should mention here that this won't work for apps that didn't follow our recommendations regarding the directories structure etc. By default, we try to move `src` and `test` folder. If one changed the original name, that would be a problem.

content/cli/workspaces.md
```diff
@@ -0,0 +1,206 @@
...
+```json
+  "compilerOptions": {
+    "webpack": true,
+    "tsConfigPath": "apps/my-project/tsconfig.app.json"
+  }
+```
+
+The `webpack` property, when true (the default), uses [webpack]() to compile and bundle the code.  This is one key difference between *workspace mode* and *single project mode*.  *Single project mode* uses `tsc` for compilation by default.  The reason for this difference is that webpack can have significant advantages in build times and in producing a single file bundling all project components together.  If you wish to generate individual files, set `webpack` to `false`, which will cause the build process to use `tsc`.
+
+**Note to Kamil**: can you please explain the significance of why `compilerOptions.tsConfigPath` is set to the default project?
```
kamilmysliwiec: > **Note to Kamil**: can you please explain the significance of why `compilerOptions.tsConfigPath` is set to the default project?

That's the path we should look for when someone calls `nest build` without passing the name of the app (like: `nest build NAMEOFTHELIBORAPP`). Same with `nest start`

content/cli/workspaces.md
```diff
@@ -0,0 +1,206 @@
...
+
+### Projects
+
+As mentioned, workspaces are composed of projects, which may be of the **application** or **library** type.  These are discussed below.
+
+#### Applications
+
+As mentioned, application-type projects, or what we might informally refer to as just "applications", are complete Nest applications that you can run and deploy.  You generate an application-type project with `nest generate app`.
+
+> info **Notice** technically, these application-type projects are called `sub-apps`.  `sub-app` is the name of the schematic you run.  The `sub-app` schematic has an alias of `app`, so you can run either `nest generate app` or `nest generate sub-app`.
```
kamilmysliwiec: `sub-app` is a private name that can change eventually (to just `g application`) so we should stick to `nest g app` in the docs

content/cli/overview.md
johnbiundo: Next update adds details on this.  Will need review to judge if it resolves this.

content/cli/workspaces.md
johnbiundo: Changing to "standard mode" and "monorepo mode" in next draft.

content/cli/overview.md
```diff
@@ -1,23 +1,97 @@
...
+
+Aside from a very few specific considerations around how the **build** process works (essentially, monorepo mode simplifies build complexities that can sometimes arise from monorepo-style project structures), and built-in [library](/cli/libraries) support, the rest of the Nest features, and this documentation, apply equally to both standard and monorepo mode project structures.  In fact, you can easily switch from standard mode to monorepo mode at any time in the future, so you can safely defer this decision while you're still learning about Nest.
+
+You can use either mode to manage multiple projects. Here's a quick summary of the differences:
+
+| Feature                                              | Standard Mode                                                      | Monorepo Mode                                              |
```
kamilmysliwiec: This table looks bad on mobile. Not sure what we can do to improve this look & feel though

content/cli/workspaces.md
```diff
@@ -0,0 +1,207 @@
...
+  <div class="item">package.json</div>
+  <div class="item">tsconfig.json</div>
+  <div class="item">tslint.json</div>
+</div>
+
+The `nest` schematic has reorganized the code - moving each **application** project under the `apps` folder, and adding a project-specific `tsconfig.app.json` file.  Our original `my-project` app has become the **default project** for the monorepo, and is now a peer with the just-added `my-app`, located under the `apps` folder.
+
+The `nest-cli.json` file contains details describing the workspace.  It now looks like this:
+
+```javascript
```
kamilmysliwiec: We should probably mention that the understanding of this file content is not very(!) important

content/cli/workspaces.md
```diff
@@ -0,0 +1,207 @@
...
+
+Notice that by default, `compilerOptions` is set as follows:
+```javascript
+"compilerOptions": {
+  "webpack": true,
+  "tsConfigPath": "apps/my-project/tsconfig.app.json"
+}
+```
+
+The `webpack` property, when true (the default), uses [webpack](https://webpack.js.org/) to compile and bundle the code.  This is one key difference between monorepo mode and standard mode.  Standard mode uses `tsc` for compilation by default.  The reason for this difference is that webpack can have significant advantages in build times and in producing a single file bundling all project components together.  If you wish to generate individual files, set `webpack` to `false`, which will cause the build process to use `tsc`.
```
kamilmysliwiec: We should add that it's possible to define an extra, custom webpack configuration in order to extend defaults (`compilerOptions.webpackConfigPath`)

Example:

```
"compilerOptions": {
  "webpack": true,
  "tsConfigPath": "apps/my-project/tsconfig.app.json",
  "webpackConfigPath": "apps/my-project/webpack.config.js"
}
```

By default, Nest CLI will look for `webpack.config.js` in the root directory.

Example `webpack.config.js` (equal to a regular webpack configuration file):

```javascript
module.exports = {
  externals: []
}
```

(above setting means that webpack will try to bundle `node_modules`, while, by default, it's excluded)

Also, the `webpack.config.js` can expose a `function` that takes default options and returns a modified object.

```javascript
export default function(options) {
   return {
      ...options,
      externals: [],
   }
}
```

content/cli/workspaces.md
```diff
@@ -0,0 +1,207 @@
...
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "declaration": false,
+    "outDir": "../../dist/apps/my-app"
+  },
+  "include": ["src/**/*"],
+  "exclude": ["node_modules", "dist", "test", "**/*spec.ts"]
+}
+```
```
kamilmysliwiec: I'm not sure if we need to show the file content here.

content/cli/overview.md
johnbiundo: I agree - MD generated tables look bad even on desktop.  I will experiment with HTML tables after we have finalized the content.

content/cli/workspaces.md
johnbiundo: Based on this, and your edit removing the `nest-cli.json` file, I decided to move things around a bit.  I think it better addresses the need to show the right level of detail on the `nest-cli.json` file, but puts it at the end so it doesn't interfere with the first-time reader getting through the main concepts.  Let me know what you think.

content/cli/workspaces.md
johnbiundo: Great, this is in the updated draft.

PR #718, opened 2019-09-30
docs(cache): update the docs to include the @CacheTTL() decorator

joeyslack:
## PR Checklist

- [ ✓ ] The commit message follows our guidelines: https://github.com/nestjs/docs.nestjs.com/blob/master/CONTRIBUTING.md

## PR Type

```
[ ] Bugfix
[ ✓ ] Feature
[ ] Build related changes
[ ✓ ] Other... Please describe: Updating docs for the CacheTTL decorator, as per: https://github.com/nestjs/nest/pull/2943
```

Issue Number: https://github.com/nestjs/nest/pull/2943

```
[ ] Yes
[✓] No
```

## Other information

### DIFF
--- a/content/techniques/caching.md
+++ b/content/techniques/caching.md
@@ -70,6 +70,7 @@ You can also apply the `CacheInterceptor` to WebSocket subscribers as well as Mi
 ```typescript
 @@filename()
 @CacheKey('events')
+@CacheTTL(20)
 @UseInterceptors(CacheInterceptor)
 @SubscribeMessage('events')
 handleEvent(client: Client, data: string[]): Observable<string[]> {
@@ -77,6 +78,7 @@ handleEvent(client: Client, data: string[]): Observable<string[]> {
 }
 @@switch
 @CacheKey('events')
+@CacheTTL(20)
 @UseInterceptors(CacheInterceptor)
 @SubscribeMessage('events')
 handleEvent(client, data) {
@@ -84,13 +86,13 @@ handleEvent(client, data) {
 }
 ```
 
-> info **Hint** The `@CacheKey()` decorator is imported from `@nestjs/common` package.
+> info **Hint** The `@CacheKey()` and `@CacheTTL()` decorators are imported from `@nestjs/common` package.
 
-However, the additional `@CacheKey()` decorator is required in order to specify a key used to subsequently store and retrieve cached data. Also, please note that you **shouldn't cache everything**. Actions which perform some business operations rather than simply querying the data should never be cached.
+However, the additional `@CacheKey()` decorator is required in order to specify a key used to subsequently store and retrieve cached data. Furthermore, the `@CacheTTL()` decorator allows setting a cache expiration time (TTL) that overrides the global cache interceptor default, allowing users to override cache durations on a per method basis. Also, please note that you **shouldn't cache everything**. Actions which perform some business operations rather than simply querying the data should never be cached.
 
 #### Customize caching
 
-All cached data has its own expiration time (TTL). To customize default values, pass the options object to the `register()` method.
+All cached data has its own expiration time (TTL) that may be overriden with the `@CacheTTL()` decorator. To customize default values, pass the options object to the `register()` method.
 
 ```typescript
 CacheModule.register({
--- a/content/techniques/caching.md
+++ b/content/techniques/caching.md
@@ -74,6 +74,26 @@ CacheModule.register({
 });
 ```
 
+#### Global cache overrides
+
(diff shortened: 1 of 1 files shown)


content/techniques/caching.md
```diff
@@ -70,27 +70,29 @@ You can also apply the `CacheInterceptor` to WebSocket subscribers as well as Mi
 ```typescript
 @@filename()
 @CacheKey('events')
+@CacheTTL(20)
```
kamilmysliwiec: I don't think we should use `@CacheTTL` by default. Instead, a separate section should be created :)

joeyslack: Fair enough. Will make.

joeyslack: I've done what you suggested. 

In addition, I moved the "Customize caching" block to be above the "Websockets & Microservices" block, as I think it makes more sense in the context of the details inside of the caching details. Since the "Customize caching" bock specifies setting cache defaults (ttl & max), it makes more sense to read this before moving onto the CacheKey & CacheTTL blocks below it. If you don't think this makes sense, I can change it back, but to me this reads better.

joeyslack:
@johnbiundo `@CacheTTL()` works anywhere that `@CacheKey()` works. In the current live documentation, CacheKey is only mentioned under WebSockets & Microservices, as that should cover every application (I think....), since it does not discriminate transport method, it should cover pretty much everything.

I removed the `@SubscribeMessage()` decorator from the example just to show how it can be used on Microservices, not only websockets (since `@SubscribeMessage()` is websocket specific). I'm more than happy to add it back, but just wanted to show the user the flexibility here.

I'd suggest that, if we were to expand on this, we'd want to expand on the `@CacheKey` documentation too, as it's not clear about this as well. I basically just wanted to keep the CacheTTL documentation as close to the CacheKey documentation as possible, since they always cover the same applicability & scope.

I'd be curious to what you think, happy to make any changes to documentations based on your feedback.

johnbiundo:
@joeyslack I wanted to mention that the documentation you provided is very good (I'm not being critical), and also that I have little to know experience with Nest's Microservices and Websockets packages, so I'm probably only in a position to comment on writing and clarity, not so much on these particular tech details :).  That said, if I understand what you're saying, this feature (along with `@CacheKey`) are relevant **exclusively** to Microservice and Websockets based apps, and **not** to HTTP based apps.  If that's true, then I think the **organization** of the chapter is fine.

Regarding the sample, I also think that's fine, though I would have a slight preference to being explicit in the docs about what the sample shows (and maybe adding that it works in both Websockets and Microservices), but that's a minor point.

I'd be happy to continue reviewing any contributions you make on this if I can help, and hopefully I'll find time to learn more about those packages so I can comment with a bit more authority :)

joeyslack:
@johnbiundo Come to think of it, the CacheTTL would apply to HTTP apps as well (unlike CacheKey), so I'll have to update this documentation regarding Websockets & Microservices. I'll send a commit for re-review.

As such, I'll update the sample to reflect these changes.

joeyslack:
@johnbiundo  Documentation has been updated for your review.

CacheKey and CacheTTL decorators apply to HTTP and non HTTP apps, so the documentation related to that is under the global overrides section.

I'm happy to make any changes you suggest.

content/techniques/caching.md
```diff
@@ -74,6 +74,26 @@ CacheModule.register({
...
+  @CacheTTL(20) // 20 second TTL
+  findAll(): string[] {
+    return [];
+  }
+}
+```
+
+> info **Hint** The `@CacheKey()` and `@CacheTTL()` decorators are imported from the `@nestjs/common` package.
+
+The `@CacheKey()` decorator may be used with or without a corresponding `@CacheTTL()` decorator and vice versa. One may choose to override only the `@CacheKey()` or only the `@CacheTTL()`. Settings that are not overriden with a decorator will use the default values as registered globally (see [Customize caching](https://docs.nestjs.com/techniques/caching#customize-caching)).
```
johnbiundo: overridden <-- slight typo

content/techniques/caching.md
```diff
@@ -95,34 +115,32 @@ handleEvent(client, data) {
...
+> info **Hint** The `@CacheTTL()` decorator may be used with or without a corresponding `@CacheKey()` decorator. 
 
 #### Different stores
 
-This service take advantage of [cache-manager](https://github.com/BryanDonovan/node-cache-manager) under the hood. The `cache-manager` package supports a wide-range of useful stores, for example, [Redis](https://github.com/dabroek/node-cache-manager-redis-store) store. A full list of supported stores is available [here](https://github.com/BryanDonovan/node-cache-manager#store-engines)). To set up the Redis store, simple pass the package together with corresponding options to the `register()` method.
+This service take advantage of [cache-manager](https://github.com/BryanDonovan/node-cache-manager) under the hood. The `cache-manager` package supports a wide-range of useful stores, for example, [Redis](https://github.com/dabroek/node-cache-manager-redis-store) store. A full list of supported stores is available [here](https://github.com/BryanDonovan/node-cache-manager#store-engines). To set up the Redis store, simple pass the package together with corresponding options to the `register()` method.
```
johnbiundo: takes advantage <-- (not take) (p.s., I know this was the original, but while you're at it :smile)

PR #737, opened 2019-10-03
Sub-Domain Routing documentation

nonplus:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] Other... Please describe:
```

Issue Number: https://github.com/nestjs/nest/pull/2980

## What is the new behavior?

Added "Sub-Domain Routing" documentation.

```
[ ] Yes
[x] No
```

### DIFF
--- a/content/controllers.md
+++ b/content/controllers.md
@@ -290,6 +290,64 @@ findOne(id) {
 }
 ```
 
+#### Sub-Domain Routing
+
+The `@Controller` decorator can take a `host` option to require that the HTTP host of the incoming requests matches some specific value. In the following example, all three routes match the same path (`/`) but two of them only respond to specific host names:
+
+```typescript
+@Controller({ host: 'admin.example.com' })
+export class AdminController {
+  @Get()
+  index(): string {
+    return "Admin page";
+  }
+}
+
+@Controller({ host: 'mobile.example.com' })
+export class MobileController {
+  @Get()
+  index(): string {
+    return "Mobile page";
+  }
+}
+
+@Controller()
+export class AppController {
+  @Get()
+  index(): string {
+    return "Default page";
+  }
+}
+```
+
+##### Host parameters
+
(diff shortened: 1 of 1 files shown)


content/controllers.md
```diff
@@ -290,6 +290,64 @@ findOne(id) {
...
+Routes are evaluated in the order in which they are configured.  This means that controllers with static `host` options should be configured first and controllers without `host` options need to be configured last:
+
+```typescript
+@Module({
+  controllers: [AdminController, AccountController, AppController],
+})
+export class AppModule {}
+```
+
+> warning **Warning** Since **Fastify** lacks support for nested routers, when using sub-domain routing, the (default) Express routing should be used instead.
```
johnbiundo: Maybe we should say Express adapter instead of Express routing.  Maybe moving (default) too?

'... when using sub-domain routing, the Express adapter (the default) should be used instead./

content/controllers.md
```diff
@@ -290,6 +290,64 @@ findOne(id) {
...
+  @Get()
+  getInfo(@HostParam('account') account: string) {
+    return account;
+  }
+}
+```
+
+##### Host matching order
+
+Routes are evaluated in the order in which they are configured.  This means that controllers with static `host` options should be configured first and controllers without `host` options need to be configured last:
```
johnbiundo: Interesting.  I did not realize this.  When you say "order in which they're configured", is that the order in which they're listed in the `controllers` metadata?  Maybe that could be spelled out a little more explicitly?

content/controllers.md
```diff
@@ -290,6 +290,64 @@ findOne(id) {
...
+// Dynamic host capturing subdomain in "account" parameter
+@Controller({ host: ':account.example.com' })
+export class AccountController { 
+  @Get()
+  getInfo(@HostParam('account') account: string) {
+    return account;
+  }
+}
+```
+
```
johnbiundo: We should probably add the standard info box about where to import `@HostParam` from.

Issue #731, opened 2019-10-02
In first steps, use `npx` for project initialization


eropple:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[ ] Feature request
[x] Documentation issue or request (new chapter/page)
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

Currently, docs.nestjs.com recommends that one install `@nestjs/cli` globally to use it for `nest new project-name`.

## Expected behavior

Recommend instead that `@nestjs/cli` be run as an `npx` command: `npx -p @nestjs/cli nest new project-name`.

## What is the motivation / use case for changing the behavior?

This can cause issues in organizations (who, even with internal docs, are likely to refer to docs.nestjs.com) where users have different globally-installed versions. Using `npx` avoids polluting the global namespace, while making it harder for users to accidentally use an unexpected version of the `nest` binstub when inside of a project that's defined `@nestjs/cli` in `devDependencies`.

kamilmysliwiec:
If we switched to `npx` by default, we'd also have to change every hint about `nest g` usage in the docs to, for example, `npx -p @nestjs/cli nest g controller app` which doesn't make any sense. I'm fine with adding a note about `npx` though (and maybe yarn as well(?)).

> This can cause issues in organizations (who, even with internal docs, are likely to refer to docs.nestjs.com) where users have different globally-installed versions.

That's why for `nest start` and `nest build` NPM scripts should be used (to use locally installed packages)

eropple:
I don't see how that follows. The change would be to change `nest g controller app` to `npx nest g controller app` or `yarn exec nest g controller app`, because once a project is created `@nestjs/cli` is in `devDependencies` (I checked). By using `npx` or `yarn exec`, you are _guaranteed_ to always be using the version specified in the project, and isn't that exactly what a user wants to do?

Over the last couple of years NodeJS seems to have standardized on this pattern and I'm not aware of many other projects where adding this has been a big problem. OTOH, I've personally stepped on the rake before of having something installed globally, forgetting `npx`, and getting unwanted behaviors out of the globally-installed binary. (Like...with NestJS. ;) )

kamilmysliwiec:
> and getting unwanted behaviors out of the globally-installed binary. (Like...with NestJS. ;) )

Just curious: what was the example unwanted behavior you got from globally-installed `@nestjs/cli`?

eropple:
[...]

PR #744, opened 2019-10-05
docs(CLI) updates

johnbiundo:
## PR Checklist

## PR Type

- [x] Documentation

## Other information

Mainly adds new `compilerOptions`.

Also addresses #731 (well, not FirstSteps, but in the CLI.  We can go back and do FirstSteps once this one is approved)

Also cleans up a few leftovers from first version.

### DIFF
--- a/content/cli/overview.md
+++ b/content/cli/overview.md
@@ -4,11 +4,9 @@ The [Nest CLI](https://github.com/nestjs/nest-cli) is a command-line interface t
 
 #### Installation
 
-Install the CLI using **npm**.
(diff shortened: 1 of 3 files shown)


kamilmysliwiec:
I'm not entirely sure if we should use **warning** quote for that:

image

It looks quite scary (like a deprecation warning 😄). Maybe we can just turn it into a normal paragraph?

johnbiundo:
Yeah, it looks a lot scarier rendered than it did in MD preview 😄 

@kamilmysliwiec Do you have any other changes?  I could re-submit changing it to an INFO note, but seems like you might prefer plain text?  If so, and you have no other comments, do you want to decide which way to go and merge it?

I'm OK either way, but my rationale for **some** call-out format is that it is really more of a "read this one time/side note" to the docs, so trying not to disrupt the main flow when someone is returning to this page for reference, as opposed to reading it for the first time.

kamilmysliwiec:
> @kamilmysliwiec Do you have any other changes? I could re-submit changing it to an INFO note, but seems like you might prefer plain text?

I think the plain text would be enough honestly. :)

Everything else looks great!

johnbiundo:
OK, changes made.  Sorry for the ugly commit history.  I had some local git corruption and had to reclone, and, well... that's what it is.  If you want me to clean that up, let me know.

johnbiundo:
Fixed!

PR #778, opened 2019-10-19
docs(logger) Edits for grammar and style, other improvements

johnbiundo:
## PR Checklist

## PR Type

- [x] Documentation

## Other information

Logger chapter edits


johnbiundo:
@kamilmysliwiec: After doing the original push, I decided to treat some more in-depth topics.  Please read and let me know if a) they're accurate 😉 and b) if you think they're appropriate.  I do see a lot of questions about customized logging, so I know it's a bit advanced, but hopefully you agree it's useful (and accurate!)

content/techniques/logger.md
```diff
@@ -65,13 +81,55 @@ export class MyLogger extends Logger {
...
+    Logger.warn('About to return cats!');
+    return this.cats;
+  }
+}
+```
+
+You can alternatively instantiate the `Logger` in your class, and take advantage of its optional constructor arguments which allow you to pass in:
+
+- `context`: a string displayed in the log output, used to identify the context of a message; see example below
+- `isTimestampEnabled`: a boolean flag that controls display of an elapsed time in the log output; defaults to false
```
kamilmysliwiec: I think we should simply remove the note about`isTimestampEnabled` here

johnbiundo:
@kamilmysliwiec I've applied your suggestions.  Should be good to go :)

content/techniques/logger.md
```diff
@@ -65,13 +81,36 @@ export class MyLogger extends Logger {
...
+
+To use the built-in logger **in your own code**, simply instantiate the `Logger` in your class, and take advantage of its optional `context` constructor argument, which allows you to pass in a string to uniquely identify the logging context. The context string will automatically be appended to your log message:
+
+```typescript
+import { Injectable, Logger } from '@nestjs/common';
+
+@Injectable()
+export class CatsService {
+  private readonly cats: Cat[] = [];
+  private readonly logger = new Logger('CatsService');
```
kamilmysliwiec: This partially solves the problem because it's still hard to mock them/disable logs when running, e.g. unit tests. In the best-case scenario, logger should be registered as a provider and injected through a constructor. We don't have to update the source code, but we should at least mention what would be the best to do here. :D

jmcdo29: Would it be suggested to create a `log` method in each class that uses the logger so it would be easy to add the context to each log call. Something like
```ts
@Injectable()
export class MyService {

  constructor(private readonly logger: Logger) {}

  private log(message: string, level: 'log' | 'error' | 'warn' | 'verbose' | 'debug') {
    switch(level) {
       case 'info': this.logger.log(message, MyService.name);
                    break;
       /* same follows for other cases with correct level */
    }
  }
}
```
Or is there some way to set the context so that it doesn't need to be sent each time? I know I solved this in my application with a TRANSIENT logger that could be injected with a dynamic module. Just wondering if there is another way you'd suggest

johnbiundo:
@kamilmysliwiec @jmcdo29 I think this update addresses all of the feedback.  Anyway, it's ready for review.  Let me know any comments.

PR #855, opened 2019-11-22
doc() No Schema without at least one query

amanDAlmond:
Inform the user that it is impossible to start generating a schema without at least one query, in case they want to start with a mutation. It's actually a very plausible thing to want to start with a mutation, a login for exemple, so this is useful to know in the getting started.

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[x] Other... Please describe: Documentation
```

While following getting started, it feels weird that you are not able to generate a schema without a query if the doc doesn't say so

Issue Number: N/A

## What is the new behavior?
The documentation now says so

```
[ ] Yes
[x] No
```

### DIFF
--- a/content/graphql/quick-start.md
+++ b/content/graphql/quick-start.md
@@ -166,7 +166,7 @@ GraphQLModule.forRoot({
 
 The `autoSchemaFile` indicates a path where your automatically generated schema will be created. Additionally, you can pass the `buildSchemaOptions` property - an options object which will be passed in to the `buildSchema()` function (from the `type-graphql` package).
 
-Please keep in mind that you have to create a query to start generating a schema, a mutation or subscription will not work.
+> info **Hint** A query must be created before schema generation takes place. A mutation or subscription will otherwise not work as expected.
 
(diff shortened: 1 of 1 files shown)


content/graphql/quick-start.md
```diff
@@ -166,6 +166,8 @@ GraphQLModule.forRoot({
 
 The `autoSchemaFile` indicates a path where your automatically generated schema will be created. Additionally, you can pass the `buildSchemaOptions` property - an options object which will be passed in to the `buildSchema()` function (from the `type-graphql` package).
 
+Please keep in mind that you have to create a query to start generating a schema, a mutation or subscription will not work.
```
prateekkathal: Maybe 😄 

```ts
> info **Hint** A query must be created before schema generation takes place. A mutation or subscription will otherwise not work as expected.
```

kamilmysliwiec, line 169:
```suggestion
> info **Hint** A query must be created before schema generation takes place. A mutation or subscription will otherwise not work as expected.
```

PR #881, opened 2019-12-12
docs(common): typed decorators through generics

galesky:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] Other... Please describe:
```

Issue Number: 
https://github.com/nestjs/nest/issues/3356
https://github.com/nestjs/nest/pull/3558

## What is the new behavior?
The user can now define type checks on Custom Decorators using generics,

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/content/custom-decorators.md
+++ b/content/custom-decorators.md
@@ -169,7 +169,7 @@ export function Auth(...roles: Role[]) {
     SetMetadata('roles', roles),
     UseGuards(AuthGuard, RolesGuard),
     ApiBearerAuth(),
-    ApiUnauthorizedResponse({ description: 'Unauthorized"' }),
+    ApiUnauthorizedResponse({ description: 'Unauthorized"' })
   );
 }
 ```
--- a/content/custom-decorators.md
+++ b/content/custom-decorators.md
@@ -108,7 +108,7 @@ Let's define a decorator that takes a property name as key, and returns the asso
 @@filename(user.decorator)
 import { createParamDecorator } from '@nestjs/common';
 
-export const User = createParamDecorator<string>((data: string, req) => {
+export const User = createParamDecorator((data: string, req) => {
   return data ? req.user && req.user[data] : req.user;
 });
 @@switch
@@ -136,7 +136,8 @@ async findOne(firstName) {
 ```
 
 You can use this same decorator with different keys to access different properties. If the `user` object is deep or complex, this can make for easier and more readable request handler implementations.
-> info **Hint** The `createParamDecorator<T>` is a generic in which `T` can be used to indicate the valid types of the data argument, if no type T is set it will assume `any`;
+
+> Info **Hint** For typescript users, note that `createParamDecorator<T>()` is a generic. This means you can explicitly enforce type safety, for example `createParamDecorator<string>((data, req) => ...)`. Alternatively, specify a parameter type in the factory function, for example `createParamDecorator((data: string, req) => ...)`. If you omit both, the type for `data` will be `any`.
 
 #### Working with pipes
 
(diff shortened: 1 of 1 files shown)


galesky:
@johnbiundo Sure, no problem! 

I'll update it on monday and will ping you here so you can help me check if we can still make some improvements. :smile:

galesky:
Hello @johnbiundo,

I've updated the docs. When you can, take a look if this is fit, I'm open for feedback (:

johnbiundo:
@galesky Thanks for taking the time to write this out.  I now have a better feel for it!  This is a nice feature for ensuring type safety in custom decorators.  

But I have a question: it appear that the correct typing is inferred properly now due to your change (which is nice!).  

So from a doc perspective, one way of looking at this (I'm going to state this as an assertion, but I'm not confident I'm correct, so please correct/challenge this!) is that there is actually no need to document anything.  In essence, what would have previously compiled properly but failed at runtime now gets properly caught as a type error at compile time (or in the IDE with intellisense).

This comes from the observation that I get the exact same behavior, whether I use
a) `export const User = createParamDecorator<string>((data: string, req) => {
  return 'john';
});`
or 
b) `export const User = createParamDecorator((data: string, req) => {
  return 'john';
});`

And then try
```typescript
@Get()
  async findAll(@User(10) name: string): Promise<Cat[]> {
    console.log('name: ', name);
    return this.catsService.findAll();
  }
```
In either case, I get the same intellisense or compile error:

> src/cats/cats.controller.ts:50:23 - error TS2345: Argument of type '10' is not assignable to parameter of type 'string | PipeTransform<any, any> | Type<PipeTransform<any, any>>'.

> 50   async findAll(@User(10) name: string): Promise<Cat[]> {
 
So, in a nutshell, I'm saying that unless I'm missing something (very possible!), I don't think there's a need to document anything.  We now just nicely enforce type safety on these custom decorators.

Your thoughts?

galesky:
Yup, you are right! On my previous commit I only hinted that you can use ecorators

galesky:
Yup, you are right! On my previous commit, I only hinted that you can use Generics and also made clear that if you don't define a type it will default to any; I would suggest we keep this version https://github.com/nestjs/docs.nestjs.com/pull/881/commits/f6fb2a711237b17b1e77f7c7bce6d70492edfdab, what do you think ?

content/custom-decorators.md
```diff
@@ -137,6 +137,55 @@ async findOne(firstName) {
...
+async wrongFindOne(@User(200) firstName: string) { // THIS FAILS! As the User decorator expects a string
+  console.log(`Hello ${firstName}`);
+}
+```
+
+If we had not defined a type, the `data` parameter would default to `any`;
+
+In practice, typing allows us to build safer decorators that can be checked while coding and on build, avoiding unexpected behavior both when implementing on `Controllers` as on coding its own functionality;
+
+The `createParamDecorator` also allows the definition of type through Generics, so the `user.decorator` can be rewritten as:
```
prateekkathal: To be honest, I'd just add this line (and the code below) after [Line#111](https://github.com/nestjs/docs.nestjs.com/pull/881/files#diff-c815b0ab017a4ee9e5d1e4170c61a3e2R111).

The rest of it kind of belongs in a blog/article somewhere, not a website documentation. 😄

johnbiundo:
Thanks @galesky and @prateekkathal .  I think we agree we simply need an info box, with no further elaboration.  My suggestion is the following (based on both of your input):

> Info **Hint** For typescript users, note that `createParamDecorator<T>()` is a generic.  This means you can explicitly enforce type safety, for example `createParamDecorator<string>((data, req) => ...)`.  Alternatively, specify a parameter type in the factory function, for example `createParamDecorator((data: string, req) => ...)`.  If you omit both, the type for `data` will be `any`.

galesky:
That's very good, thanks @johnbiundo, i'll update this PR later today

galesky:
Hello @johnbiundo , I've updated the docs to match what we have discussed here.

PR #923, opened 2020-01-07
fix(): remove smooth scroll

yannbertrand:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] Other... Please describe:
```

Searching something in the docs using browser search is really inconvenient as the user has to wait a few seconds for the page to scroll.

Kapture 2020-01-07 at 16 30 48

## What is the new behavior?

Directly scroll when typing

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/src/styles.scss
+++ b/src/styles.scss
@@ -4,10 +4,6 @@
 @import './scss/variables.scss';
 @import './scss/utils.scss';
 
-:root {
-  scroll-behavior: smooth;
-}
-
 body {
   font-size: 16px;
   font-weight: 400;


lazarljubenovic:
To be honest I was weighing if I should report this is a bug in Chrome or here, and never got around to do either, so forgot about it. Now seeing that it's confusing for other people as well, maybe it means that it's indeed Chrome's fault for doing this. 

I'm wondering if it should be reported to webkit or Chrome, though? 🤔

yannbertrand:
I didn't think about it but it does make sense!

If you want to get a normal scroll behavior on Nest's docs until then you can use:
```js
document.documentElement.style.setProperty('scroll-behavior', 'auto')
```

BrunnerLivio:
Its a [chrome bug](https://support.google.com/chrome/thread/23003371).
I think temporarily setting `auto` scroll behavior as @yannbertrand mentioned is the best option for now.

[Dan Abramov confirmed it too on Twitter](https://twitter.com/dan_abramov/status/1214605699625504770?s=21)

kamilmysliwiec:
Let's close this then

yannbertrand:
I would have preferred this temporary fix integrated until it's solved on Chrome's side. As you can see on the gif from the original post, it's really annoying for Chrome users, and we have no idea on when they will resolve this.

PR #931, opened 2020-01-11
docs(cli): Add docs for new generateOptions conf

egodsk:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] Other... Please describe:
```

Issue Number: N/A

## What is the new behavior?
Adds docs for the new generateOptions, as seen in PR https://github.com/nestjs/nest-cli/pull/527

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/content/cli/workspaces.md
+++ b/content/cli/workspaces.md
@@ -175,6 +175,7 @@ The top-level properties are as follows:
 - `"collection"`: points at the collection of schematics used to generate components; you generally should not change this value
 - `"sourceRoot"`: points at the root of the source code for the single project in standard mode structures, or the _default project_ in monorepo mode structures
 - `"compilerOptions"`: a map with keys specifying compiler options and values specifying the option setting; see details below
+- `"generateOptions"`: a map with keys specifying global generate options and values specifying the option setting; see details below
 - `"monorepo"`: (monorepo only) for a monorepo mode structure, this value is always `true`
 - `"root"`: (monorepo only) points at the project root of the _default project_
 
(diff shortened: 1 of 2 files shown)


content/cli/workspaces.md
```diff
@@ -190,6 +191,76 @@ These properties specify the compiler to use as well as various options that aff
 | `deleteOutDir`      | boolean             | If `true`, whenever the compiler is invoked, it will first remove the compilation output directory (as configured in `tsconfig.json`, where the default is `./dist`).                                                                    |
 | `assets`            | array               | Enables automatically distributing non-TypeScript assets whenever a compilation step begins (asset distribution does **not** happen on incremental compiles in `--watch` mode).  See below for details.                                  |
 
+#### Global generate options
+
+These properties specify the default generate options for when running the `nest generate` command.
```
johnbiundo: Making a few suggestions.  The following is the first, and is the format I will use.

"for when running" -> "to be used by"

Above means I suggest replacing existing text (left of arrow) with the suggestion to the right of the arrow.

content/cli/workspaces.md
```diff
@@ -190,6 +191,76 @@ These properties specify the compiler to use as well as various options that aff
...
+
+| Property Name       | Property Value Type | Description                                                                                                                                                                                                                              |
+| ------------------- | ------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `spec`              | boolean *or* object | If assigned to a boolean it can be specified as either `true` or `false` to respectively enable or disable `spec` generation by default, if not specified otherwise in the cli, or in the specific project `generateOptions`. If assigned to an object, each key represents a schematic name, and the boolean value represents if the default spec generation is enabled / disabled for that specific schematic. |
```
johnbiundo: "If assigned to a boolean it can be specified as either `true` or `false` to respectively enable or disable `spec` generation by default, if not specified otherwise in the cli, or in the specific project `generateOptions`." -> "If the value is boolean, a value of `true` enables `spec` generation and a value of `false` disables it.  A flag passed on the CLI command line overrides this setting, as does a project-specific `generateOptions` setting."

johnbiundo: "If assigned to an object, each key represents a schematic name, and the boolean value represents if" -> "If the value is an object, each key represents a schematic name, and the boolean value determines whether"

content/cli/workspaces.md
```diff
@@ -190,6 +191,76 @@ These properties specify the compiler to use as well as various options that aff
...
+
+Below is given an example of global generate options, both with spec being a boolean and as an object
```
johnbiundo: "Below is given an example of global generate options, both with spec being a boolean and as an object" ->

"The following example uses a boolean value to specify that `spec` file generation should be disabled by default for all projects:"

content/cli/workspaces.md
```diff
@@ -190,6 +191,76 @@ These properties specify the compiler to use as well as various options that aff
...
+  "generateOptions": {
+    "spec": {
+      "service": false
+    }
+  },
+  ...
+}
+```
+
+> error **Warning** Specifying the `spec` as an object, the key for the generation schematic does currently not support automatic alias handling. This means that specifying a key as for example `service: false` and trying to generate a service via the alias `s`, the spec would still be generated. To make sure both the normal schematic name and the alias works, specify both the normal command name as well as the alias, as seen below.
```
johnbiundo: "Specifying the `spec` " -> "When Specifying the `spec` "

"does currently not" -> "does not currently"

"the alias works" -> "the alias work as intended'

content/cli/workspaces.md
```diff
@@ -190,6 +191,76 @@ These properties specify the compiler to use as well as various options that aff
...
+>       "s": false
+>     }
+>   },
+>   ...
+> }
+> ```
+
+#### Project specific generate options
+
+As well as being able to write global generate options, project specific generate options is also available. The project specific generate options follow the exact same format as the global generate options, but are specified directly on each project.
```
johnbiundo: "As well as being able to write global generate options, project specific generate options is also available."  -> "In addition to providing global generate options, you may also specify project-specific generate options."

content/cli/workspaces.md
```diff
@@ -190,6 +191,76 @@ These properties specify the compiler to use as well as various options that aff
...
+
+Project specific generate options overwrite global generate options.
```
johnbiundo: overwrite -> override

content/cli/workspaces.md
```diff
@@ -190,6 +191,76 @@ These properties specify the compiler to use as well as various options that aff
...
+        }
+      },
+      ...
+    }
+  },
+  ...
+}
+```
+
+> notice **Notice** Project specific generate options overwrite global generate options, meaning that any options set on the project specific generate options take a high precedence than options set in the global generate options. Directly specifying generate specific options on the CLI take the highest precedence than both project specific options and global generate options.
```
johnbiundo: "Project specific generate options overwrite global generate options, meaning that any options set on the project specific generate options take a high precedence than options set in the global generate options. Directly specifying generate specific options on the CLI take the highest precedence than both project specific options and global generate options."

->

"The order of precedence for generate options is as follows.  Options specified on the CLI command line take precedence over project-specific options.  Project-specific options override global options."

content/cli/workspaces.md
```diff
@@ -190,6 +191,76 @@ These properties specify the compiler to use as well as various options that aff
...
+```javascript
+{
+  "generateOptions": {
+    "spec": false
+  },
+  ...
+}
+```
+
+or
```
johnbiundo: "or" -> "In this example, `spec` file generation is disabled only for `service` schematics (e.g., `nest generate service...`):

egodsk:
Hi @johnbiundo!

Thank you for the great feedback! I have committed the changes to my fork. They all looked great to me :)

PR #936, opened 2020-01-13
docs(common): remove return await syntax

jmcdo29:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[x] Code style update (formatting, local variables)
[ ] Build related changes
[ ] Other... Please describe:
```

Currently, there are 37 locations across 11 files in the docs that use `return await`, which is [considered a bad practice](https://eslint.org/docs/rules/no-return-await). 

## What is the new behavior?

No use of `return await` in the docs

```
[ ] Yes
[x] No
```

## Other information

Search and replace was done by VSCode.

### DIFF
--- a/content/graphql/guards-interceptors.md
+++ b/content/graphql/guards-interceptors.md
@@ -12,7 +12,7 @@ You can use either [guards](/guards), [interceptors](/interceptors), [filters](/
 @Query('author')
 @UseGuards(AuthGuard)
 async getAuthor(@Args('id', ParseIntPipe) id: number) {
-  return await this.authorsService.findOneById(id);
+  return this.authorsService.findOneById(id);
 }
 ```
 
@@ -22,7 +22,7 @@ As you can see, GraphQL works pretty well with both guards and pipes. Thanks to 
 @Mutation()
 @UseInterceptors(EventsInterceptor)
 async upvotePost(@Args('postId') postId: number) {
-  return await this.postsService.upvoteById({ id: postId });
+  return this.postsService.upvoteById({ id: postId });
 }
 ```
 
--- a/content/graphql/mutations.md
+++ b/content/graphql/mutations.md
@@ -16,17 +16,17 @@ export class AuthorResolver {
 
   @Query('author')
   async getAuthor(@Args('id') id: number) {
-    return await this.authorsService.findOneById(id);
+    return this.authorsService.findOneById(id);
   }
 
   @Mutation()
   async upvotePost(@Args('postId') postId: number) {
-    return await this.postsService.upvoteById({ id: postId });
+    return this.postsService.upvoteById({ id: postId });
   }
 
   @ResolveProperty('posts')
   async getPosts(@Parent() { id }) {
-    return await this.postsService.findAll({ authorId: id });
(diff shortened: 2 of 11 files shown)


kamilmysliwiec:
If we don't use `await` in those methods anymore, `async` is not needed either.

Anyways, sometimes in the docs, we use `async` + `await` (even when not required) just to emphasize that `async` methods are widely supported. Hence, I'm not sure if we should remove `async` from method signatures or not.

jmcdo29:
Okay, maybe instead of removing them we can make sure it is clear that using `return await` is not suggested and that it is only done to demonstrate that asynchronous methods are widely supported? Though I'm not sure where in the docs that would go... If you don't want to keep discussing this, feel free to close the PR. I get your point about showing the availability of async functions in Nest.

kamilmysliwiec:
Maybe we should revert `resolvers-map.md`, `pipes.md`, and `guards.md` (since their role is to introduce new features), but leave the rest (especially in recipes, etc)

jmcdo29:
Sounds like a plan. Should we also add in a note after those three saying we are only using the `return await` syntax to show NestJS works with promises and otherwise this is usually a bad pattern?

kamilmysliwiec:
> otherwise this is usually a bad pattern?

I wouldn't say that we should call it a bad pattern. `return await` is (usually) a redundant instruction, typically wiped out by tools like tslint/eslint (depending on the configuration). It's still valid though and doesn't lead to side-effects. In some cases (try..catch blocks), `return await` is even required so such warning (blockquote) in the docs may cause some confusion.

lazarljubenovic:
I've seen `return await` being called an anti-pattern so many times, but still fail to see why. It emphasizes that the method returns a promise in a clear and concise way that's easy to refactor if you need to (eg. by doing `res = await` before returning).

It's not necessary to put it there, and it CAN mean that you don't know how it works, but that's not a reason to call it anti-pattern. You can also drop `{` around a single statement when writing `if`'s blocks, and nobody's getting rid of those - in fact, they're encouraged.

jmcdo29:
According to the `eslint` docs (linked above), it is considered an anti-pattern because the return of an async function is always wrapped in a promise, thus awaiting it could add time before the final promise resolution. To my understanding `return await` would be similar to doing something like `Promise.resolve(Promise.resolve(someAsyncCode))` which _usually_ is not helpful.

kamilmysliwiec:
@jmcdo29 I think this depends on the compiler (if we downgrade the code) + V8 runtime. For the latest versions (and we recommend using them), this shouldn't change anything.

jmcdo29:
[There was a recent PR](https://github.com/nestjs/nest/pull/3723) to the main repository that removed this syntax from the sample classes. I don't believe it ended up breaking anything, so I figured the docs should be in line as well. This PR has generated a lot more conversation than I thought it would, so do you think we should go ahead with removing the syntax from all but the mentioned files, or go ahead and keep it and let people ask questions?

kamilmysliwiec:
Alright, let's just remove `await` and leave `async` there (to inform people that `async` methods are supported)

PR #948, opened 2020-01-22
docs(task-scheduling): add a short note about CronExpression enum

jeffminsungkim:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[x] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] Other... Please describe:
```

Issue Number: https://github.com/nestjs/schedule/pull/33

## What is the new behavior?
Supply `CronExpression` enumerations

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/content/techniques/task-scheduling.md
+++ b/content/techniques/task-scheduling.md
@@ -99,6 +99,25 @@ Some sample cron patterns are:
   </tbody>
 </table>
 
+The `@nestjs/schedule` package provides a convenience enum with commonly used cron patterns. You can use this enum as follows:
+
+```typescript
+import { Injectable, Logger } from '@nestjs/common';
+import { Cron, CronExpression } from '@nestjs/schedule';
+
+@Injectable()
+export class TasksService {
+  private readonly logger = new Logger(TasksService.name);
+
+  @Cron(CronExpression.EVERY_45_SECONDS)
+  handleCron() {
+    this.logger.debug('Called every 45 seconds');
+  }
+}
+```
+
+In this example, the `handleCron()` method will be called every `45` seconds.
+
 Alternatively, you can supply a JavaScript `Date` object to the `@Cron()` decorator. Doing so causes the job to execute exactly once, at the specified date.
 
 > info **Hint** Use JavaScript date arithmetic to schedule jobs relative to the current date. For example, `@Cron(new Date(Date.now() + 10 * 1000))` to schedule a job to run 10 seconds after the app starts.


kamilmysliwiec:
Thanks for your contribution @jeffminsungkim! I think we shouldn't remove the example with a plain string though (`* * * * * 45`). This new enum is useful, but still, we want to emphasize that passing a string is fine too (gives more control)

jeffminsungkim:
> Thanks for your contribution @jeffminsungkim! I think we shouldn't remove the example with a plain string though (`* * * * * 45`). This new enum is useful, but still, we want to emphasize that passing a string is fine too (gives more control)

It sounds reasonable. I'll fix it. 😺

content/techniques/task-scheduling.md
```diff
@@ -53,13 +53,32 @@ export class TasksService {
 
 In this example, the `handleCron()` method will be called each time the current second is `45`. In other words, the method will be run once per minute, at the 45 second mark.
 
+Some of you might consider defining a simple cron pattern with a plain string is a cumbersome task to be done. You can use the `CronExpression` enum instead, which provides commonly used cron patterns, as follows:
```
johnbiundo: I have a few suggestions.

1. For this paragraph, I suggest the following rewording

From: "Some of you might consider defining a simple cron pattern with a plain string is a cumbersome task to be done. You can use the `CronExpression` enum instead, which provides commonly used cron patterns, as follows:"

To: "The @nestjs/schedule package provides a convenience enum with commonly used cron patterns.  You can use this enum as follows:"

2. I think the section flows better if we move this documentation to the bottom of the section, just  after the section with the samples (Starting with "Some sample cron patterns are" and ending with the last sample).

**Question**: Is there an external reference where these are documented?  (I.e., is this a well known set of patterns, or is this something you created from scratch?).  The reason I'm asking is that if there is a good external reference, we should cite it here to help people know the available enum values.  It is probably too much to add these directly into the doc, so I'd rather do a link if one exists.  If not, that's fine too. :)

content/techniques/task-scheduling.md
```diff
@@ -53,13 +53,32 @@ export class TasksService {
...
+In this example, the `handleCron()` method will be called every `45` seconds.
+
 The `@Cron()` decorator supports all standard [cron patterns](http://crontab.org/):
 
 - Asterisk (e.g. `*`)
 - Ranges (e.g. `1-3,5`)
 - Steps (e.g. `*/2`)
 
-In the example above, we passed `45 * * * * *` to the decorator. The following key shows how each position in the cron pattern string is interpreted:
+In the first example above, we passed `45 * * * * *` to the decorator. The following key shows how each position in the cron pattern string is interpreted:
```
johnbiundo: I think if we make the change above (moving to the bottom), this change can be reverted.

content/techniques/task-scheduling.md
jeffminsungkim: @johnbiundo Although I've thought carefully about the section flows before deciding where to put it, the idea wasn't good enough. 😆 (Also, I like your phrase! I'll revise it.)

In response to your question, finding references for commonly used cron expressions was difficult. So I've written down the patterns that came to my mind one by one, which considered being simple and frequently being used.

Just an FYI, I didn't write complex patterns on purpose.

jeffminsungkim:
> I made a few suggestions.

@johnbiundo I followed your suggestions and fixed it.

kamilmysliwiec:
Thanks!!

PR #961, opened 2020-01-29
fix link for injection scopes

vologab:
Logger section contains relative link to fundamentals/injection-scopes, but we need root link, because otherwise it points to https://docs.nestjs.com/techniques/logger/fundamentals/injection-scopes


kamilmysliwiec:
Thank you!

PR #986, opened 2020-02-07
improvement(docs): another argument name

iliyaZelenko:
It looks a little strange that the argument is called `id`, although in fact it is a class that contains different properties, one of which is `id`.


kamilmysliwiec:
Thanks!

PR #995, opened 2020-02-09
docs(mongo): mention need for extending Cat interface

MrOnlineCoder:
The following line will create not very user friendly compilation errors because of the incorrect Cat interface declaration:
```
constructor(@InjectModel('Cat') private readonly catModel: Model<Cat>) {}
```
The `Cats` example is used almost everywhere through the documentation, but it's not stated anywhere in the docs that you have to `extend` the interface from the mongoose `Document`. Surely, it can be easily fixed by looking at the official example, but that step is not quite straightforward, and, in my opinion, should be mentioned explicitly.

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[ ] No
```

## Other information

### DIFF
--- a/content/techniques/mongo.md
+++ b/content/techniques/mongo.md
@@ -58,7 +58,7 @@ import { CatSchema } from './schemas/cat.schema';
 export class CatsModule {}
 ```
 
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
Since we've changed the way you can create your mongoose models, this PR became outdated. Thanks for your contribution though!

PR #1038, opened 2020-02-26
Fix a typo - HeroService 👉🏼 HeroController

the6thm0nth:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] Other... Please describe:
```

Fix a typo on docs about microservices/grpc

Issue Number: N/A

## What is the new behavior?
`HeroService` to `HeroController`

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/content/microservices/grpc.md
+++ b/content/microservices/grpc.md
@@ -177,7 +177,7 @@ You can also omit the first `@GrpcMethod()` argument. In this case, Nest automat
 ```typescript
 @@filename(hero.controller)
 @Controller()
-export class HeroService {
+export class HeroController {
   @GrpcMethod()
   findOne(data: HeroById, metadata: any): Hero {
     const items = [


kamilmysliwiec:
This one is tricky. So in fact, gRPC handlers have to be registered as controllers, but conventionally they are known as "gRPC services". This is a nomenclature used and recommended by gRPC creators.

>  In this case, Nest automatically associates the handler with the service definition from the proto definitions file based on the **class** name where the handler is defined. For example, in the following code, class `HeroService` associates its handler methods with the `HeroService` service definition in the `hero.proto` file based on the matching of the name `'HeroService'`.

the6thm0nth:
Yes, I know.

there are 2 reasons I think this is a typo
1. the file name is `hero.controller.ts`
2. there are already `HeroController` mentioned above two cases of same examples.

PR #1046, opened 2020-02-28
Update circular-dependency.md in favor of issue #4168

JustDoItSascha:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[ ] No
```

## Other information

### DIFF
--- a/content/fundamentals/circular-dependency.md
+++ b/content/fundamentals/circular-dependency.md
@@ -52,7 +52,7 @@ export class CommonService {
 }
 ```
 
-> warning **Warning** The order of instantiation is indeterminate. Make sure your code does not depend on which constructor is called first.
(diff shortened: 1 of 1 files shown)


content/fundamentals/circular-dependency.md
```diff
@@ -52,7 +52,7 @@ export class CommonService {
 }
 ```
 
-> warning **Warning** The order of instantiation is indeterminate. Make sure your code does not depend on which constructor is called first.
+> warning **Warning** The order of instantiation is indeterminate. Make sure your code does not depend on which constructor is called first. When you need to perform some actions in the constructor, you will get problems in conjunction with `forwardRef()`, better use the [Lifecycle events](/fundamentals/circular-dependency) instead.
```
johnbiundo: I propose using the following wording if I'm understanding the behavior correctly:

In cases where you have circular dependencies, it is best practice to limit the constructor's role to **only performing dependency injection**.  You should defer any other initialization work to the [lifecycle hook methods](/fundamentals/lifecycle-events) to ensure the entire application dependency tree has been constructed (including resolving circular dependencies) before further initialization commences.

PR #1082, opened 2020-03-17
docs(graphql): suggested improvements

johnbiundo:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[ ] No
```

## Other information

@kamilmysliwiec - this is WIP as I work through more testing of GraphQL.  I'll comment back when I think it's ready for review/merge.

### DIFF
--- a/content/graphql/quick-start.md
+++ b/content/graphql/quick-start.md
@@ -12,7 +12,7 @@ Start by installing the required packages:
 $ npm i --save @nestjs/graphql graphql-tools graphql
 ```
 
-And depending on what you use (Express or Fastify), you need to install `apollo-server-express` or `apollo-server-fastify`.
+Depending on what underlying platform you use (Express or Fastify), you must also install either `apollo-server-express` or `apollo-server-fastify`.
 
 #### Overview
 
@@ -20,7 +20,7 @@ Nest offers two ways of building GraphQL applications, the **code first** and th
 
 In the **code first** approach, you use decorators and TypeScript classes to generate the corresponding GraphQL schema. This approach is useful if you prefer to work exclusively with TypeScript and avoid context switching between language syntaxes.
 
-In the **schema first** approach, the source of truth is GraphQL SDL (Schema Definition Language) files. SDL is a language-agnostic way to share schema files between different platforms. Nest automatically generates your TypeScript definitions based on the GraphQL schemas (using either classes or interfaces) to reduce redundancy.
+In the **schema first** approach, the source of truth is GraphQL SDL (Schema Definition Language) files. SDL is a language-agnostic way to share schema files between different platforms. Nest automatically generates your TypeScript definitions (using either classes or interfaces) based on the GraphQL schemas to reduce the need to write redundant boilerplate code.
 
 #### Getting started with GraphQL & TypeScript
 
@@ -105,7 +105,7 @@ A fully working code first sample is available [here](https://github.com/nestjs/
 
 #### Schema first
 
-To use the schema first approach, start by adding a `typePaths` property to the options object.
+To use the schema first approach, start by adding a `typePaths` property to the options object. The `typePaths` property indicates where the `GraphQLModule` should look for GraphQL SDL schema definition files you'll be writing. These files will be combined in memory; this allows you to split your schemas into several files and locate them near their resolvers.
 
 ```typescript
 GraphQLModule.forRoot({
@@ -113,9 +113,7 @@ GraphQLModule.forRoot({
 }),
 ```
 
-The `typePaths` property indicates where the `GraphQLModule` should look for GraphQL SDL schema definition files you'll be writing. These files will be combined in memory; this allows you to split your schemas into several files and locate them near their resolvers.
-
-You will typically also need to have TypeScript definitions (classes and interfaces) that correspond to the GraphQL SDL types. Creating the corresponding TypeScript definitions by hand is redundant and tedious. It leaves us without a single source of truth -- each change made within SDL forces us to adjust TypeScript definitions as well. To address this, the `@nestjs/graphql` package can automatically generate TypeScript definitions from the abstract syntax tree ([AST](https://en.wikipedia.org/wiki/Abstract_syntax_tree)). To enable this feature, add the `definitions` options property when configuring the `GraphQLModule`.
+You will typically also need to have TypeScript definitions (classes and interfaces) that correspond to the GraphQL SDL types. Creating the corresponding TypeScript definitions by hand is redundant and tedious. It leaves us without a single source of truth -- each change made within SDL forces us to adjust TypeScript definitions as well. To address this, the `@nestjs/graphql` package can **automatically generate** TypeScript definitions from the abstract syntax tree ([AST](https://en.wikipedia.org/wiki/Abstract_syntax_tree)). To enable this feature, add the `definitions` options property when configuring the `GraphQLModule`.
 
 ```typescript
 GraphQLModule.forRoot({
--- a/content/graphql/quick-start.md
+++ b/content/graphql/quick-start.md
@@ -20,7 +20,7 @@ Nest offers two ways of building GraphQL applications, the **code first** and th
 
 In the **code first** approach, you use decorators and TypeScript classes to generate the corresponding GraphQL schema. This approach is useful if you prefer to work exclusively with TypeScript and avoid context switching between language syntaxes.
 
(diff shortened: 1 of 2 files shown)


content/graphql/quick-start.md
```diff
@@ -54,16 +54,18 @@ import { GraphQLModule } from '@nestjs/graphql';
...
 
 As mentioned, these options will be forwarded to the `ApolloServer` constructor.
 
 <app-banner-enterprise></app-banner-enterprise>
 
-#### Playground
+#### GraphQL playground
 
-The playground is a graphical, interactive, in-browser GraphQL IDE, available by default on the same URL as the GraphQL server itself. With your application running in the background, open your web browser and navigate to `http://localhost:3000/graphql` (host and port may vary depending on your configuration). You should see the GraphQL playground, as shown below.
+The playground is a graphical, interactive, in-browser GraphQL IDE, available by default on the same URL as the GraphQL server itself. To access the playground, you need a basic GraphQL server configured and running. To see it now, you can install and build the [working example here](https://github.com/nestjs/nest/tree/master/sample/12-graphql-schema-first). Alternatively, if you're following along with these code samples, once you've complete the steps in the [Resolvers chapter](/graphql/resolvers-map), you can access the playground.
```
kamilmysliwiec: > [working example here](https://github.com/nestjs/nest/tree/master/sample/12-graphql-schema-first)

Can we point to the code first example instead?

content/graphql/resolvers-map.md
```diff
@@ -1,12 +1,25 @@
...
+If we were using the schema first approach, we'd define such a schema with SDL like this:
+
+```graphql
+type Author {
+  id: Int!
+  firstName: String
+  lastName: String
+  posts: [Post]
+}
+```
```
kamilmysliwiec:
```
If we were using the schema first approach, we'd define such a schema with SDL like this:

```graphql
type Author {
  id: Int!
  firstName: String
  lastName: String
  posts: [Post]
}
```

I'm not fully convinced that showing the schema first approach as part of this section will not cause additional confusion (some people may not be even familiar with it yet). It's up to you though. If you think it's valuable, let's leave this as it is.

content/graphql/resolvers-map.md
```diff
@@ -28,11 +41,13 @@ export class Author {
 }
 ```
 
-> info **Hint** TypeScript's metadata reflection system has several limitations which make it impossible to, for instance, determine what properties a class consists of or recognize whether a given property is optional or required. Thus, we must either use the `@Field` decorator or use a [CLI plugin](/graphql/resolvers#cli-plugin).
```
kamilmysliwiec: Side topic - do you think we should add `@filename()` directives to code snippets? For instance, for `Author` we could use `author.entity` (or model) etc

johnbiundo: Yes, I definitely do!  I think it would help a **lot** to provide some guidance on how to organize code in files.  Would you be able to do that (I'm not 100% sure I can make the best recommendations)!

content/graphql/resolvers-map.md
```diff
@@ -150,23 +196,30 @@ export class AuthorResolver {
...
 }
 ```
 
-Usually you won't have to pass an object into the `@Args()` decorator as with the `getAuthor()` method above. For example, if an identifier's type is string, the following construction would be sufficient:
+The `@Query()` decorator's options object (where we pass `{{ '{' }}name: 'author'{{ '}' }}` above) accepts a number of keys:
+
+- name: name of the query; a `string`
+- description: a description that will be used to generate GraphQL schema documentation (e.g., in GraphQL playground); a `string`
+- deprecationReason: sets query metadata to show the query as deprecated (e.g., in GraphQL playground); a `string`
+- nullable: whether the query can return a null data response; `boolean` or `'items'` or `'itemsAndList'` (see above for details of `'items'` and `'itemsAndList'`)
```
kamilmysliwiec:
```suggestion
-`name`: name of the query; a `string`
- `description`: a description that will be used to generate GraphQL schema documentation (e.g., in GraphQL playground); a `string`
- `deprecationReason`: sets query metadata to show the query as deprecated (e.g., in GraphQL playground); a `string`
- `nullable`: whether the query can return a null data response; `boolean` or `'items'` or `'itemsAndList'` (see above for details of `'items'` and `'itemsAndList'`)
```

content/graphql/resolvers-map.md
```diff
@@ -175,22 +228,22 @@ getAuthor(
...
 
-With inline `@Args()` calls, the code becomes bloated. Instead, you can create a dedicated `GetAuthorArgs` class:
+With inline `@Args()` calls, code like the example above becomes bloated. Instead, you can create a dedicated `GetAuthorArgs` arguments class and access it in the handler method as follows:
 
 ```typescript
 @Args() args: GetAuthorArgs
 ```
 
-With the following body:
+Create the `GetAuthorArgs()` class using `@ArgsType()` as shown below:
```
kamilmysliwiec:
```suggestion
Create the `GetAuthorArgs` class using `@ArgsType()` as shown below:
```

content/graphql/resolvers-map.md
johnbiundo: I too am a little conflicted.  My assumption is that people who are using this will have some famliarity with GraphQL, and therefore will relate well to seeing SDL as opposed to having to make that "mapping" on their own.  But that's just a guess.  I'll ponder some more.

content/graphql/resolvers-map.md
```diff
@@ -350,9 +426,9 @@ export class CreatePostInput extends Post {
 }
 ```
 
-#### Decorators
+#### GraphQL resolver argument decorators
```
kamilmysliwiec: Maybe just `Arguments decorators` or `GraphQL argument decorators`? Since they can be used for subscriptions, queries, mutations, field resolvers etc

content/graphql/resolvers-map.md
```diff
@@ -396,10 +476,10 @@ Once we're done with the above steps, we have to provide the `AuthorResolver` so
 export class AuthorsModule {}
 ```
 
-The `GraphQLModule` will take care of reflecting the metadata and transforming classes into the correct resolvers map automatically. The only thing you need to be aware of is that you need to import this module somewhere, so Nest will be able to utilize `AuthorsModule`.
-
 > info **Hint** Learn more about GraphQL queries [here](http://graphql.org/learn/queries/).
```
kamilmysliwiec: We should probably move this hint elsewhere (maybe even duplicate in the code first and schema first sections(?))

content/graphql/resolvers-map.md
kamilmysliwiec: Perfect 💪 Example structure:

```
// AUTHOR MODULE (`authors` directory)
- dto
   - get-author.args.ts
- models
   - author.model.ts
authors.module.ts
authors.service.ts
authors.resolver.ts

// POST MODULE (`posts` directory)
- dto
   - create-post.input.ts
- models
   - post.model.ts
posts.module.ts
posts.service.ts
```

Conventionally:

- unions: `NAME.union.ts`
- enums: `NAME.enum.ts`
- plugins: `NAME.plugin.ts`
- directives: `NAME.directive.ts`

And as for interfaces, I think we should use `.model.ts` suffix as well (to avoid confusion with TS interfaces).

content/graphql/resolvers-map.md
johnbiundo: On balance, my opinion is to leave it in.  As I said, I think it's a safe assumption that people coming here are either familiar with SDL or are following basic GraphQL tutorials and are more likely to understand models as expressed by SDL (which is visibly easy to comprehend) and make the mapping from that to code-first Nest patterns.

content/graphql/resolvers-map.md
johnbiundo: Done for resolvers.  Will have to refer back to this as I go through other chapters.

johnbiundo:
@kamilmysliwiec I think this is ready for a final review.  I think I'll open a separate PR for any other GraphQL chapters, so if you are good with this, feel free to merge.

johnbiundo:
Also, let me know if you prefer I clean up the commit history.  Sorry, I normally do but did not here yet.

kamilmysliwiec:
Thank you ❤️

PR #1090, opened 2020-03-20
Incorrect types in Custom Decorator example

GrayStrider:
It may have something to do with my particular combination of guards though, but when I tried to use example from docs to get req.user, I got an `ctx.switchToHttp() is not a function` error. I took a look at the actual object in debugger and it appears to be `Request` (or `Express.Request`, in my case; if that's indeed the case by default I'll update the changes)

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ x ] Other... Please describe: Documentation
```

```
[ ] Yes
[ x ] No
```

### DIFF
--- a/content/custom-decorators.md
+++ b/content/custom-decorators.md
@@ -68,9 +68,8 @@ In order to make your code more readable and transparent, you can create a `@Use
 import { createParamDecorator, ExecutionContext } from '@nestjs/common';
 
 export const User = createParamDecorator(
-  (data: unknown, ctx: ExecutionContext) => {
-    const request = ctx.switchToHttp().getRequest();
-    return request.user;
+  (data: never, { user }: Request) => {
+    return user;
   },
 );
 ```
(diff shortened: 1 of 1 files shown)


content/custom-decorators.md
```diff
@@ -68,9 +68,8 @@ In order to make your code more readable and transparent, you can create a `@Use
 import { createParamDecorator, ExecutionContext } from '@nestjs/common';
 
 export const User = createParamDecorator(
-  (data: unknown, ctx: ExecutionContext) => {
-    const request = ctx.switchToHttp().getRequest();
-    return request.user;
+  (data: never, { user }: Request) => {
```
GrayStrider: `never` since `data` is unused here

content/custom-decorators.md
```diff
@@ -112,20 +111,14 @@ Let's define a decorator that takes a property name as key, and returns the asso
 import { createParamDecorator, ExecutionContext } from '@nestjs/common';
 
 export const User = createParamDecorator(
-  (data: string, ctx: ExecutionContext) => {
-    const request = ctx.switchToHttp().getRequest();
-    const user = request.user;
-
+  (data: keyof Request['user'], { user }: Request) => {
```
GrayStrider: In case with Passport we actually have Request.User type (which is empty and required augmentation), but this works as well

GrayStrider:
Here's a suggestion that seem to be inline with what I'm talking about:
https://github.com/nestjs/nest/issues/830#issuecomment-453939539

kamilmysliwiec:
Please, note that since 7.0.0 release, `createParamDecorator` no longer takes `Request` as a second parameter. https://docs.nestjs.com/migration-guide#custom-route-decorators

PR #1105, opened 2020-03-24
Possible typo

GrayStrider:
>...becomes hard to manually test **the** end-to-end behavior

Definite article seems extraneous in this case, since we're not talking about some _specific_ set of e2e tests, but "behaviour in general".
https://ell.stackexchange.com/a/59911/99911
But, perhaps it's due to `..each` that is later used to refer to the tests?

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ x ] Other... Please describe: documentation typo
```
```
[ ] Yes
[ x ] No
```

### DIFF
--- a/content/fundamentals/unit-testing.md
+++ b/content/fundamentals/unit-testing.md
@@ -162,7 +162,7 @@ Instead of using the production version of any provider, you can override it wit
 
 #### End-to-end testing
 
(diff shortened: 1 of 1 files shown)


johnbiundo:
Thanks for the suggestion.  I think I mildly disagree, but it's probably not clear cut.  My technique in these cases is to read the sentence without the adjective/clause and see how it "feels".  e.g., if I read aloud:

"it becomes harder to test behavior of each API endpoint", (omitted "end-to-end") that doesn't sound as natural to me as "it becomes harder to test **the** behavior of each API endpoint".

PR #1113, opened 2020-03-26
docs(pipes): add exception code in validation pipe

Dominic-Preap:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] Other... Please describe:
```

Issue Number: https://github.com/nestjs/nest/issues/4393

## What is the new behavior?
Add `errorHttpStatusCode` in `ValidationPipe` or any built-in pipes to throw a different HTTP status code

```ts
new ValidationPipe({
  errorHttpStatusCode: HttpStatus.UNPROCESSABLE_ENTITY
});

// Before
{
  "statusCode": 400,
  "message": ["email must be an email"],
  "error": "Bad Request"
}

// After
{
  "statusCode": 422,
  "message": ["email must be an email"],
  "error": "Unprocessable Entity"
}
```

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/content/pipes.md
+++ b/content/pipes.md
@@ -505,6 +505,11 @@ In addition to these, all `class-validator` options (inherited from the `Validat
     <td>If set to true, validation errors will not be returned to the client.</td>
   </tr>
   <tr>
+    <td><code>exceptionCode</code></td>
+    <td><code>number</code></td>
+    <td>Takes an HTTP status code and returns an exception object to be thrown. (Default: <code>400</code> or <code>HttpStatus.BAD_REQUEST</code>)</td>
+  </tr>
+  <tr>
     <td><code>exceptionFactory</code></td>
     <td><code>Function</code></td>
     <td>Takes an array of the validation errors and returns an exception object to be thrown.</td>
--- a/content/pipes.md
+++ b/content/pipes.md
@@ -507,7 +507,7 @@ In addition to these, all `class-validator` options (inherited from the `Validat
   <tr>
     <td><code>errorHttpStatusCode</code></td>
     <td><code>number</code></td>
-    <td>Takes an HTTP status code and returns an exception object to be thrown. (Default: <code>400</code> or <code>HttpStatus.BAD_REQUEST</code>)</td>
+    <td>This setting allows you to specify which exception type will be used in case of an error (e.g., if you pass 404, it would call <code>throw new NotFoundException()</code>)</td>
   </tr>
   <tr>
     <td><code>exceptionFactory</code></td>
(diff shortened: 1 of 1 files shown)


content/pipes.md
```diff
@@ -504,6 +504,11 @@ In addition to these, all `class-validator` options (inherited from the `Validat
     <td><code>boolean</code></td>
     <td>If set to true, validation errors will not be returned to the client.</td>
   </tr>
+  <tr>
+    <td><code>exceptionCode</code></td>
```
kamilmysliwiec: `exceptionCode` -> `errorHttpStatusCode`

Dominic-Preap: Oop!! already updated.

content/pipes.md
```diff
@@ -504,6 +504,11 @@ In addition to these, all `class-validator` options (inherited from the `Validat
...
+    <td><code>number</code></td>
+    <td>Takes an HTTP status code and returns an exception object to be thrown. (Default: <code>400</code> or <code>HttpStatus.BAD_REQUEST</code>)</td>
```
johnbiundo: Without studying this in detail, it appears to me that this description describes a function, not a setting.  Does it actually "take a value" and "return an object", or does it just cause `class-validator` to behave this way?

Also (kind of a nit), but does this option belong inside a table described (and linked to) a list of `class-validator` options?  Don't have a perfect solution for that, but it might be confusing if the user looks for more info on this option that is "nest only" in the class validator docs.

kamilmysliwiec: This setting allows you to specify which exception type will be used in case of an error (e.g., if you pass 404, it would call `throw new NotFoundException()`)

johnbiundo: Great.  I think we should adopt/adapt @kamilmysliwiec 's description above instead of the current one, which to me, reads like a description of a factory function.

Dominic-Preap: | Option                | Type     | Description|
| --------------------- | -------- | ---------- |
| `errorHttpStatusCode` | `number` | ​This setting allows you to specify which exception type will be used in case of an error (e.g., if you pass 404, it would call `throw new NotFoundException()`)|
| `errorHttpStatusCode` | `number` | ​This setting allows you to specify which exception type will be used in case of an error. By default, it would throw `new BadRequestException()`)|

@johnbiundo should we tell the user by default it will throw `BadRequestException()`? If it's not necessary, I will follow @kamilmysliwiec description.

johnbiundo: I like your second alternative.  

Suggestion: 

"By default, it would throw `new BadRequestException()`" 
-> 
"By default it throws `BadRequestException`"

Dominic-Preap: @johnbiundo it is done.

kamilmysliwiec: one minor suggestion @Dominic-Preap. Can you add `.` (dot) at the end of the sentence? :) (to stay consistent with other rows)

Dominic-Preap: it is done. 👍

kamilmysliwiec:
Thank you @Dominic-Preap!

PR #1118, opened 2020-03-27
docs: add info for gql argumentshost

johannesschobel:
Dear @kamilmysliwiec ,

this is the requested PR (see https://github.com/nestjs/graphql/issues/705 ) to update the docs and describe that you now can check for `argumentsHost.getType() === 'graphql'`.

All the best,
Johannes


johannesschobel:
Dear @johnbiundo ,
please note that this is specifically for the `ArgumentsHost` and **not** the `ExecutionContext`. Within the `ExecutionContext` you can persue as you wrote. However, as @kamilmysliwiec pointed out in the linked issue ( https://github.com/nestjs/graphql/issues/705 ), for the `ArgumentsHost` you need to use it like this.

All the best

kamilmysliwiec:
Using type arguments should be preferred (instead of explicitly casting to `string`). Please note `getType()` method is a generic method (defaults to the `ContextType` type). 

Thus, if you use `@nestjs/graphql`, there are 2 ways:
1. Recommended: use `GqlContextType` as in @johannesschobel PR
2. If you don't need type checking - use `getType<string>()`

PR #1128, opened 2020-04-01
Update module-reference.md

MickL:
I tried to make the point about instantiating providers dynamically more clear, as I struggled with it.

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[X] Other... Please describe: Updated documentation
```

### DIFF
--- a/content/fundamentals/module-reference.md
+++ b/content/fundamentals/module-reference.md
@@ -191,17 +191,32 @@ const catsRepository = await this.moduleRef.resolve(CatsRepository, contextId);
 
 #### Instantiating classes dynamically
 
-To dynamically instantiate a class that wasn't previously registered as a provider, use the module reference's `create()` method.
+To dynamically instantiate a class that wasn't previously registered as a provider but depends on other providers use the module reference's `create()` method.
+
+```typescript
+@@filename(cat)
+@Injectable({scope: Scope.TRANSIENT})
+export class Cat {
+  private name: string;
+
+  constructor(private logger: Logger) {}
+  
+  init(name: string) {
+    this.name = name;
+  }
+}
+```
 
 ```typescript
 @@filename(cats.service)
 @Injectable()
 export class CatsService implements OnModuleInit {
-  private catsFactory: CatsFactory;
+  private myCat: Cat;
   constructor(private moduleRef: ModuleRef) {}
 
   async onModuleInit() {
-    this.catsFactory = await this.moduleRef.create(CatsFactory);
+    this.myCat = await this.moduleRef.create(Cat);
+    this.myCat.init('Meowth');
   }
 }
 @@switch
@@ -213,7 +228,8 @@ export class CatsService {
   }
 
   async onModuleInit() {
-    this.catsFactory = await this.moduleRef.create(CatsFactory);
+    this.myCat = await this.moduleRef.create(Cat);
+    this.myCat.init('Meowth');
   }
 }
 ```


content/fundamentals/module-reference.md
```diff
@@ -191,17 +191,32 @@ const catsRepository = await this.moduleRef.resolve(CatsRepository, contextId);
 
 #### Instantiating classes dynamically
 
-To dynamically instantiate a class that wasn't previously registered as a provider, use the module reference's `create()` method.
+To dynamically instantiate a class that wasn't previously registered as a provider but depends on other providers use the module reference's `create()` method.
```
jmcdo29, line 194: Do these other providers need to be available in the current module's scope (i.e. Logger needs to be in the `providers` array)?

johnbiundo:
I don't know the answer to @jmcdo29 's question; @kamilmysliwiec do you have a comment on that?

@MickL I'm not sure I understand this any better with your example.  Could you please elaborate in this discussion thread what the core issue is (i.e., your statement of the need that this feature is fulfilling, and how your example shows that)?  I'm struggling to connect the dots.  If you can do that in this discussion thread, maybe we can make some improvements to the actual doc.  Thanks!

MickL:
The main issue was that I wanted to create a class dynamically e.g. `new Cat('Meowth');` but it should have the dependency injection so I can have existing services in there like the logger. The answer doing so was found in the documentation, but it was not clear to me. I wanted to alter it so others will find the informations doing so more easy.

content/fundamentals/module-reference.md
kamilmysliwiec: @jmcdo29 Yes

content/fundamentals/module-reference.md
```diff
@@ -191,17 +191,32 @@ const catsRepository = await this.moduleRef.resolve(CatsRepository, contextId);
...
 @Injectable()
 export class CatsService implements OnModuleInit {
-  private catsFactory: CatsFactory;
+  private myCat: Cat;
   constructor(private moduleRef: ModuleRef) {}
 
   async onModuleInit() {
-    this.catsFactory = await this.moduleRef.create(CatsFactory);
+    this.myCat = await this.moduleRef.create(Cat);
+    this.myCat.init('Meowth');
```
kamilmysliwiec, line 219: Can we revert this change? With the text update you proposed, everything should be clear now

kamilmysliwiec:
Just replied @johnbiundo

content/fundamentals/module-reference.md
jmcdo29: I had a feeling that was the case. It should probably be explicitly mentioned if this is going to move forward.

johnbiundo:
@jmcdo29 can you propose a final update to this?  I'm a little unclear after the back and forth where you guys think this should end up.  I can take a final pass on any style/grammar.

content/fundamentals/module-reference.md
MickL: Creating a factory was kind of confusing to me. I think it should be clear that when you can create a class, you can also create it using a factory. Or am I wrong?

jmcdo29:
@johnbiundo I'll see if I have some time soon to propose some changes. This PR and #1129 kind of go hand in hand with how some of the parts work, so I'll try to help coordinate that

MickL:
@jmcdo29 I appreciate that. For me it was very hard to find the right words but I hope you get what I wanted to archive with both PRs.

johnbiundo:
@jmcdo29 Thanks.  I posted a suggestion on #1129 earlier as well.  No substantive changes, just style.

content/fundamentals/module-reference.md
kamilmysliwiec: `Factory` acts as a random (sample) name here. We can change it to, let's say, `CatsHelpers` or `CatsStorage`, or `CatsUtils`. I refer to the rest of the additional code (which seems unnecessary to me)

adelin-b:
A suggestion for making it even more useful would be showing how to `this.moduleRef.create(Service)` and passing argument to the Service.constructor

PR #1152, opened 2020-04-10
docs(subscription.md) correction re: subscription handler name

johnbiundo:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[ ] No
```

## Other information

Partially addresses https://github.com/nestjs/graphql/issues/753 by correcting an error and being more explicit about how the published event needs to correspond to the subscription result object.

Doesn't address the [mutation variant](https://github.com/nestjs/graphql/issues/753#issuecomment-611945956) of that issue as that seems like a corner case, and when I tried to explain it, I felt it was more confusing than helpful.

### DIFF
--- a/content/graphql/subscriptions.md
+++ b/content/graphql/subscriptions.md
@@ -45,6 +45,21 @@ type Subscription {
 }
 ```
 
+Note that subscriptions, by definition, return an object with a single top level property whose key is the name of the subscription. This name is either inherited from the name of the subscription handler method (i.e., `commentAdded` above), or is provided explicitly by passing an option with the key `name` as the second argument to the `@Subscription()` decorator, as shown below.
+
+```typescript
+  @Subscription(returns => Comment, {
+    name: 'commentAdded',
+  })
+  addCommentHandler() {
+    return pubSub.asyncIterator('commentAdded');
+  }
+```
+
+This construct produces the same SDL as the previous code sample, but allows us to decouple the method name from the subscription.
+
+### Publishing
+
 Now, to publish the event, we use the `PubSub#publish` method. This is often used within a mutation to trigger a client-side update when a part of the object graph has changed. For example:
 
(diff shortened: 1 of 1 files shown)


josh18:
Looks good :+1:

Rather than
> The above requirement can create undesired coupling between the `@Subscription()` method handler name (`commentAdded()`) and the event payload. To specify a name for the subscription that is independent of the method handler name, use the `name` option in the `@Subscription()` decorator:

I wonder if it would be clearer to note that the payload key must match the GraphQL Subscription name? If you don't define this name explicitly using the name option, then the name will be inherited from the name of the subscription handler. Just so it is obvious that they are two different concepts.

> Doesn't address the mutation variant of that issue as that seems like a corner case, and when I tried to explain it, I felt it was more confusing than helpful.

Are you talking about the resolve option? If so I think all that we need to say is that if you are using the resolve option you must return the unwrapped payload.

❌ 
```
@Subscription(returns => Comment, {
  resolve: payload=> payload,
})
commentAdded() { ... }
```

✔️
```
@Subscription(returns => Comment, {
  resolve: payload=> payload.commentAdded,
})
commentAdded() { ... }
```
We might want to update some of the examples to reflect that as well

johnbiundo:
@josh18 Take a look.  I think this is an improvement -- though not exactly what you suggested.

I didn't notice your comment about examples until after I pushed this.  Gotta run for now, but take a look and I can make final changes tomorrow. Thanks for your help.

josh18:
Yup I think that's good :+1:

johnbiundo:
@kamilmysliwiec I think this is ready for your review.

kamilmysliwiec:
Thank you!

PR #1167, opened 2020-04-17
Update graphql/mapped-types.md

longngn:
The second argument to PickType/OmitType/PartialType should be the decorator factory of the derived class not the base class.

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[x] Other... Please describe: Docs fix
```

Issue Number: N/A

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/content/graphql/mapped-types.md
+++ b/content/graphql/mapped-types.md
@@ -35,7 +35,7 @@ export class UpdateUserInput extends PartialType(CreateUserInput) {}
 
 > info **Hint** The `PartialType()` function is imported from the `@nestjs/graphql` package.
 
-The `PartialType()` function takes an optional second argument that is a reference to the decorator factory of the child class. In the example above, we are extending `CreateUserInput` which is annotated with the `@InputType()` decorator. We didn't need to pass `InputType` as the second argument since the child class is also annotated with `@InputType()`. If you want to extend a class decorated with `@ObjectType`, pass `InputType` as the second argument. For example:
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
Thanks @longngn, good catch!

content/graphql/mapped-types.md
```diff
@@ -35,11 +35,11 @@ export class UpdateUserInput extends PartialType(CreateUserInput) {}
 
 > info **Hint** The `PartialType()` function is imported from the `@nestjs/graphql` package.
 
-The `PartialType()` function takes an optional second argument that is a reference to the decorator factory of the type being extended. In the example above, we are extending `CreateUserInput` which is annotated with the `@InputType()` decorator. We didn't need to pass `InputType` as the second argument since it's the default value. If you want to extend a class decorated with `@ObjectType`, pass `ObjectType` as the second argument. For example:
+The `PartialType()` function takes an optional second argument that is a reference to the decorator factory of the child class. In the example above, we are extending `CreateUserInput` which is annotated with the `@InputType()` decorator. We didn't need to pass `InputType` as the second argument since the child class is also annotated with `@InputType()`. If you want to extend a class decorated with `@ObjectType`, pass `InputType` as the second argument. For example:
```
johnbiundo, line 38:
```suggestion
The `PartialType()` function takes an optional second argument that is a reference to a decorator factory.  This argument can be used to change the decorator function applied to the resulting (*child*) class.  If not specified, the child class effectively uses the same decorator as the *parent* class (the class referenced in the first argument).  In the example above, we are extending `CreateUserInput` which is annotated with the `@InputType()` decorator.  Since we want `UpdateUserInput` to also be treated as if it were decorated with `@InputType()`, we didn't need to pass `InputType` as the second argument. If the parent and child types are different, (e.g., the parent is decorated with `@ObjectType`), we would  pass `InputType` as the second argument. For example:
```

johnbiundo:
I made some suggestions on wording changes.  A little more verbose, but (perhaps?) a little more clear.  It's a slightly nuanced subject, so I think it makes sense to be more explicit with the explanation.

In any case, let me know if you think that works, or feel free to suggest alternatives.  Let's get it right :)

Once we have the right pattern, I can apply it to the other functions in this chapter as well.

longngn:
@johnbiundo Thanks for the suggestions! I'm not a native speaker so I could see that being explicit and redundant in documentation is clearer than being concise.

johnbiundo:
@kamilmysliwiec Mind taking a quick look before I replicate this to the other sections?

@longngn Thanks for your contribution!

kamilmysliwiec:
Looks great @johnbiundo :)

PR #969, opened 2020-02-02
[Refactor] TabsComponent to CodeComponent custom element 

BrunnerLivio:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/docs.nestjs.com/blob/master/CONTRIBUTING.md

## PR Type
What kind of change does this PR introduce?

```
[ ] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[x] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] Other... Please describe:
```

## What is the current behavior?
In the future, the HTML shall be appended to the DOM dynamically (pre-compile Markdown to HTML using Dgeni, dynamically request the HTML page). The `TabsComponent` is an Angular component - and the easiest way to say to Angular "hey, in this HTML I dynamically loaded is a component", is to simply convert the Angular component to a Custom Element / Webcomponent. Angular.io uses the same technique.
 
Related Issue Number: #549

## What is the new behavior?
Simplified the interface of the code/switcher component & converted it to a custom element.

```html
<code-element filename="cat.service">
  <pre slot="ts" class="language-typescript">
    <code class="language-typescript">
console.log("Hello");
    </code>
  </pre>
  <pre slot="js" class="language-javascript">
    <code class="language-javascript">
console.log("Hello");
    </code>
  </pre>
</code-element>
```

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

## Other information
The `[appAnchor]` directive will need to be converted to a native element to enable #549. This will be done in a follow-up PR.

PR #1214, opened 2020-05-02
Reapply code component: fix not displaying code when only filename is set

BrunnerLivio:
## PR Checklist

## PR Type

```
[x] Bugfix
[x] Feature
[ ] Build related changes
[ ] Other... Please describe:
```

Follow-up of issue #969

```
[ ] Yes
[ ] No
```

## Other information


johnbiundo:
The one I think I've run into was "code + filename" without switch.  I think you're saying above that this was broken and is now fixed?  I've found this one to be most important, as if you are writing language-independent samples and want to include a filename (which I think is usually useful to help indicate code/directory conventions) it failed.  I might be misremembering because I think I stopped using filename without switch quite a while ago, and chalked this up to some issue that we didn't have any control over.

Short answer: Yay!  Thanks Livio!

BrunnerLivio:
> I think you're saying above that this was broken and is now fixed?

No the Example "Code + Switch" without filename was and is still broken :P

```typescript
console.log('typescript')
@@switch
console.log('javascript')
```
Results into two elements (supposed to be one)

Selection_113

But like I said, this was broken before too. The behavior with this PR should be identical. I added a reference for master branch in my previous comment.

I am fine with enabling this as well, but maybe in a separate PR :)

johnbiundo:
@BrunnerLivio I think the case I was referring to was "code + filename".  That's the one I was avoiding because at one point, I think it did not work.  I may be wrong about that, but I have avoided "code + filename" because I thought it was broken.  Your example shows that working (and maybe it was working before, but in any case, that's news to me!) that I'm most interested in!

kamilmysliwiec:
@BrunnerLivio 

see some regressions:

image

https://deploy-preview-1214--docs-nestjs.netlify.app/pipes#validation-use-case

image

https://deploy-preview-1214--docs-nestjs.netlify.app/pipes#object-schema-validation

(see "filename" bar)

PR #1218, opened 2020-05-05
Improve lifecycle's documentation

ScreamZ:
## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ x ] Documentation improvement
```

```
[ ] Yes
[ x ] No
```
----
I just added a small paragraph to show that those hooks won't be called as long you are not registering hooks. I had to try on my own because it was not explicit enough at the beginning for me 😅

### DIFF
--- a/content/fundamentals/lifecycle-events.md
+++ b/content/fundamentals/lifecycle-events.md
@@ -24,7 +24,7 @@ In the following table, `onModuleDestroy`, `beforeApplicationShutdown` and `onAp
 | `beforeApplicationShutdown()`\* | Called after all `onModuleDestroy()` handlers have completed (Promises resolved or rejected);<br />once complete (Promises resolved or rejected), all existing connections will be closed (`app.close()` called). |
 | `onApplicationShutdown()`\*     | Called after connections close (`app.close()` resolves.                                                                                                                                                           |
 
-\* For those events, if you're not calling `app.close()` explicitly, you must opt-in to make them work with system's signals: see `Shutdown application` section below.
+\* For these events, if you're not calling `app.close()` explicitly, you must opt-in to make them work with system signals such as `SIGTERM`.  See [Application shutdown](#application-shutdown) below.
 
 #### Usage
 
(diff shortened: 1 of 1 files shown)


ScreamZ:
Thanks for feedback John,
Sounds good to me, but we should maybe add a `*` (asterisk) to the events in the table with something like as reference:

```* For those events, you must opt-in to make them work: see `Shutdown application` section below.```

Because if you don't read the full page, you might be surprised that they are not enabled by default, and it's not that expensive in terms of characters adding :p 
And then we can add what you wrote in the end ?

Sound good? I can edit the PR that way if you agree @johnbiundo

**Comment by johnbiundo** [comment]
@ScreamZ I think that works!  If you want to go ahead and make that change, that would be great.  Thanks!

**Comment by ScreamZ** [comment]
@johnbiundo Sound good now, I let you check, just rebased etc…

One small question about the repo? There is no hot reload feature while changing markdown content, is it normal? 

Have a nice day

File: content/fundamentals/lifecycle-events.md [hunk 1]
```diff
@@ -14,13 +14,17 @@ The following diagram depicts the sequence of key application lifecycle events,
...
+| `onApplicationBootstrap()`      | Called once all modules have been initialized, but before listening for connections.                                                                                                                              |
+| `onModuleDestroy()`\*           | Called after a termination signal (e.g., `SIGTERM`) has been received.                                                                                                                                            |
+| `beforeApplicationShutdown()`\* | Called after all `onModuleDestroy()` handlers have completed (Promises resolved or rejected);<br />once complete (Promises resolved or rejected), all existing connections will be closed (`app.close()` called). |
+| `onApplicationShutdown()`\*     | Called after connections close (`app.close()` resolves.                                                                                                                                                           |
+
+\* For those events, if you're not calling `app.close()` explicitly, you must opt-in to make them work with system's signals: see `Shutdown application` section below.
```
johnbiundo, line 27:
```suggestion
\* For these events, if you're not calling `app.close()` explicitly, you must opt-in to make them work with system signals such as `SIGTERM`.  See [Application shutdown](#application-shutdown) below.
```

content/fundamentals/lifecycle-events.md
```diff
@@ -64,28 +68,30 @@ async onModuleInit() {
...
+The `onModuleDestroy()`, `beforeApplicationShutdown()` and `onApplicationShutdown()` hooks are called in the terminating phase (in response to an explicit call to `app.close()` or upon receipt of system signals such as SIGTERM if opted-in). This feature is often used with [Kubernetes](https://kubernetes.io/) to manage containers' lifecycles, by [Heroku](https://www.heroku.com/) for dynos or similar services.
 
-To use these hooks you must activate a listener which listens to shutdown signals.
+Disabled by default for performances optimizations, to use these hooks **you must enable them** by calling `enableShutdownHooks()`, which listens to shutdown signals:
```
johnbiundo, line 73:
```suggestion
Shutdown hook listeners consume system resources, so they are disabled by default. To use shutdown hooks, you **must enable listeners** by calling `enableShutdownHooks()`:
```

ScreamZ:
@johnbiundo 

Applied, good job mate :)

Feel free to merge squash :)

See you soon :)

kamilmysliwiec:
Thanks!

PR #1264, opened 2020-05-27
Add a small note to the Transactions section

matias-pg:
To run a transaction we need to inject the `Connection` object into a service, but to keep the database-related code inside a (custom) repository we need to inject the `EntityManager`, since custom repositories are being instantiated by `typeorm` and not by `@nestjs/core` and therefore can't use the NestJS dependency injection.

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[x] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] Other... Please describe:
```

Issue Number: [#493 (at @nestjs/typeorm)](https://github.com/nestjs/typeorm/issues/493)

## What is the new behavior?
Mention that custom repositories need to use the `EntityManager` object-

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/content/techniques/sql.md
+++ b/content/techniques/sql.md
@@ -364,7 +364,7 @@ A database transaction symbolizes a unit of work performed within a database man
 
 There are many different strategies to handle [TypeORM transactions](https://typeorm.io/#/transactions). We recommend using the `QueryRunner` class because it gives full control over the transaction.
 
-First, we need to inject the `Connection` object into a class in the normal way:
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
>  (in custom repositories we need to inject `EntityManager`):

Custom repositories don't support injection at all. The `EntityManager` is passed in by the `typeorm` library (as a sole argument)

matias-pg:
I just realized that it isn't even necessary to inject `EntityManager` lol, in custom repositories it can be accessed via `this.manager`.

Sorry man, I should've read a bit more the documentation.

PR #1351, opened 2020-07-12
docs(microservices) clarify applicability of @MessagePattern decorator

robgee86:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[x] Other... Please describe: docs update
```

Issue Number: N/A

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/content/microservices/basics.md
+++ b/content/microservices/basics.md
@@ -97,7 +97,7 @@ The request-response message style is useful when you need to **exchange** messa
 
 To enable the request-response message type, Nest creates two logical channels - one is responsible for transferring the data while the other waits for incoming responses. For some underlying transports, such as [NATS](https://nats.io/), this dual-channel support is provided out-of-the-box. For others, Nest compensates by manually creating separate channels. There can be overhead for this, so if you do not require a request-response message style, you should consider using the event-based method.
 
-To create a message handler based on the request-response paradigm use the `@MessagePattern()` decorator, which is imported from the `@nestjs/microservices` package. This decorator is only registered for [controllers](https://docs.nestjs.com/controllers) since they are the entry points for your application: it would be bad practice to distribute `@MessagePattern()` decorators throughout the application code.
+To create a message handler based on the request-response paradigm use the `@MessagePattern()` decorator, which is imported from the `@nestjs/microservices` package. This decorator should be used only within the [controller](https://docs.nestjs.com/controllers) classes since they are the entry points for your application. Using them inside providers won't have any effect as they are simply ignored by Nest runtime.
 
 ```typescript
 @@filename(math.controller)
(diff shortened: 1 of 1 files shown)


content/microservices/basics.md
```diff
@@ -97,7 +97,7 @@ The request-response message style is useful when you need to **exchange** messa
 
 To enable the request-response message type, Nest creates two logical channels - one is responsible for transferring the data while the other waits for incoming responses. For some underlying transports, such as [NATS](https://nats.io/), this dual-channel support is provided out-of-the-box. For others, Nest compensates by manually creating separate channels. There can be overhead for this, so if you do not require a request-response message style, you should consider using the event-based method.
 
-To create a message handler based on the request-response paradigm use the `@MessagePattern()` decorator, which is imported from the `@nestjs/microservices` package.
+To create a message handler based on the request-response paradigm use the `@MessagePattern()` decorator, which is imported from the `@nestjs/microservices` package. This decorator is only registered for [controllers](https://docs.nestjs.com/controllers) since they are the entry points for your application: it would be bad practice to distribute `@MessagePattern()` decorators throughout the application code.
```
johnbiundo, line 100: Are we saying:
a) you should not have @MessagePattern() decorators throughout your code (e.g., in non-controller classes) because it's bad practice

or 

b) you can not do so (and Nest enforces this).

I believe the answer is (a), but I don't think the wording is completely clear on this.

@kamilmysliwiec @robgee86

robgee86: I meant both, in particular point (b). I'm new to NestJS and while quickly trying the features it offers I tried to put @MessagePattern decorator in services but couldn't trigger the decorated method. I didn't find any documentation about this anti-pattern so I've decided to create the pull request to clarify that this decorator should and must be used in controllers.
Should I rephrase maybe?
Thanks.

johnbiundo: @kamilmysliwiec could you weigh in please?  

@robgee86 What I meant to ask by point (b) is that I don't think (I haven't tested this, and maybe you can clarify based on your experience doing so) Nest warns you (either in TypeScript or at run time) that this pattern won't work. I'm assuming it just silently fails.  Assuming that's the case, I'd say we should go with (a) as the main point, and strengthen the advice that you *should not* use @MessagePattern anywhere other than in controllers.

robgee86: Yes, exactly, there's no warning and the decorated method is never triggered so it is silently "failing". I'll rephrase and strengthen point (a) but I think it should also be pointed out that using the decorator outside of controllers is simply not allowed.

kamilmysliwiec: There's no warning.

>  I think it should also be pointed out that using the decorator outside of controllers is simply not allowed.

Fully agree with @robgee86

kamilmysliwiec:
Would you mind looking at (grammar wise etc)? :) @jmcdo29

jmcdo29 (approved):
Grammatically the change looks great. Only thing to add would be the suggestion John made of making sure it's known that `@MessagePattern()` _technically_ can be set in files that aren't controllers, but that Nest won't pick those decorated methods up, but that can be a future enhancement if needed

content/microservices/basics.md
kamilmysliwiec, line 100:
```suggestion
To create a message handler based on the request-response paradigm use the `@MessagePattern()` decorator, which is imported from the `@nestjs/microservices` package. This decorator should be used only within the [controller](https://docs.nestjs.com/controllers) classes since they are the entry points for your application. Using them inside providers won't have any effect as they are simply ignored by Nest runtime.
```

PR #1376, opened 2020-07-25
docs(openapi): add generics api response to openapi other features

nartc:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[ ] No
```

## Other information


kerrmarin:
This is a very cool workaround but how does this allow strongly typed languages that are consumers of an API in nestjs to generate their clients? There is ambiguity in the resulting types used by `allOf` because there is no type information associated, simply "an object" that will have the properties described but not _what_ object that is.

This works really well for validation when you already have an object, but when you are trying to generate an API client for a strongly-typed language I don't see how this would work. Is this

nartc:
@kerrmarin I totally understand the concern. However, this is TypeScript limitation because we **CANNOT** get metadata of generics at runtime natively.

As far as client code generation, I don't think it's going to be a big deal. I'm going to list an example. Assume I have the following model:

```
{
   results: [],
   limit: 10,
   offset: 0,
   total: 100
}
```

As first glance, most people will realize this is a model of a `PaginatedResult` that has some information about the current pagination state. In C#, the model is a class as followed:

```csharp
    public class PaginatedResult<T>
    {
        public PaginatedResult(IEnumerable<T> list, long totalCount, long offset, long limit)
        {
            Results = new List<T>(list ?? new T[0]);
            Total = totalCount;
            Limit = limit;
            Offset = offset;
        }

        [Required]
        [Description("Total number of records in query")]
        public long Total { get; private set; }

        [Required]
        [Description("Collection of results")]
        public IReadOnlyList<T> Results { get; private set; }

        [Required]
        [Description("The number of records skipped")]
        public long Offset { get; private set; }

        [Required]
        [Description("Number of records returned")]
        public long Limit { get; private set; }

    }
```
Annotate an endpoint in .NET:

```csharp
[SwaggerResponse(HttpStatusCode.OK, typeof(LimitedResult<SomeModel>))]
```
C# will be able to get metadata of `LimitedResult<SomeModel>` because of its strong Reflection capability. The end-result of the Swagger schema is: `LimitedResultOfSomeModel` as its own schema.

```
"oneOf": [
   {
      "$ref": "#/components/schemas/LimitedResultOfSomeModel"
   }
]
```

With TypeScript, we are going to have to deal with:

```
"allOf": [
   {
      "$ref": "#/components/schemas/PaginatedResult"
   },
  {
      "properties": {
         "results": {
            "$ref": "#/components/schemas/SomeModel"
         }
      }
   }
]
```

Which isn't the end of the world. The generated client stub would still make a lot of sense imho.

kerrmarin:
@nartc yeah I think that's my point. Using 'allOf' means that client generation isn't possible because there is ambiguity (i.e what _concrete_ class should be generated from that swagger def?) I guess what I am concerned about is offering this as an 'official' recommendation because it doesn't actually solve the, imho, main use case of OpenAPI: client generation.

nartc:
@kerrmarin I strongly believe that will be the responsibility of the client generation tools. `allOf` is OpenAPI 3.0's official solution to handle Inheritance. 

About ambiguity, depending on what generation tool you're using, I'm using https://github.com/RicoSuter/NSwag and here's a workaround to get rid of the ambiguity: 

image
Add a `title` property to the schema with `PaginatedResponseOf${model.name}`, so if I have `PaginatedResponse<UserVm>`, I will end up with: `title: PaginatedResponseOfUserVm`. The generated client will have the following models (TypeScript):

image
image

Would this be enough to mention in the docs? If you think of a better way, please feel free to raise a feature request and/or PR.

kerrmarin:
Yes, I'm not arguing that it's not the generation tool's job to generate the type, but if it doesn't have enough information (i.e. there is ambiguity in the JSON) no tool can be expected to just "make it up".

The workaround you propose looks promising, I'll take a closer look and run some tests on the generator we're using (https://github.com/OpenAPITools/openapi-generator in case you're curious).

Thanks for addressing my concerns! I think adding the workaround for ambiguity will be a great addition to this PR

kamilmysliwiec:
I'm not sure if the "Other features" chapter is the best choice for this guide. What do you think about putting this into the "Operations" instead? We could also add a prefix - "Advanced:" to the section title.

nartc:
- Move to Operations instead of Other Features.
- Address ambiguity that @kerrmarin pointed out.

kamilmysliwiec:
Thanks!

PR #1431, opened 2020-08-24
Better suggestion about fastify rate limiter

ozkanonur:


kamilmysliwiec:
There are plenty of NestJS community packages that wrap the official libraries, but since it's really hard to track & review all of them, in the docs we're trying to refer to the officially maintained packages

PR #1441, opened 2020-08-29
Improve docs about using Logger

mayorandrew:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] Other... Please describe:
```

I was trying to integrate [`nestjs-pino`](https://github.com/iamolegga/nestjs-pino) (json logger) in my app in the best possible way, and struggled a little bit since nestjs docs don't quite clearly explain how to best use custom logger in your app services.

The example with transient scopes seems strange to me because why would I inject custom logger to my services if I am already setting it as globally default using `app.useLogger`. If I am switching loggers, I would have to do search and replace across all application. Also, I've noticed that when I use default `Logger` from `@nestjs/common` it actually uses the custom logger from `app.useLogger`. This was not clear to me, so I started digging.

I found this amazing Stack Overflow answer https://stackoverflow.com/a/52907695/4601673 but didn't quite get how Nest can use custom logger from `app.useLogger` if I am instantiating `Logger` from `@nestjs/common` in my class. 

So I digged through the source. If we look how `app.useLogger` is implemented (https://github.com/nestjs/nest/blob/67e15afbf33364ac6dcf0ade58d26aade7c2eb1f/packages/core/nest-application-context.ts#L123), it actually calls static method `Logger.overrideLogger` of the `Logger` from `@nestjs/common` which changes the static property `Logger.instance` (https://github.com/nestjs/nest/blob/01dfe29458a2851e42efe8d5b4d614caf9935742/packages/common/services/logger.service.ts#L70).

Now, if we look, how `Logger` itself is implemented, whenever you call instance methods like `log`/`warn`, it actually calls the related method from the saved `Logger.instance` (https://github.com/nestjs/nest/blob/01dfe29458a2851e42efe8d5b4d614caf9935742/packages/common/services/logger.service.ts#L108).

Basically, by calling `app.useLogger` we set the logger which will be used through the default `Logger` instances from `@nestjs/common`. So there is no need to inject custom logger if we only use methods from `LoggerService` interface.

If we dig more into nestjs source code, the logger is actually used the same way, as provided in the SO anwser linked above, so I think that is how it should be. As I see it, nestjs itself and possible third-party libraries would just use `Logger` from `@nestjs/common` and assume its implementation agnosticity. And application developer could substitute any custom logger they want via `app.useLogger`.

Here are some examples of `new Logger` usage from nestjs repo: https://github.com/nestjs/nest/blob/fc05b244a8a0db15ef60e612c9e2afe0290f995a/packages/core/errors/exception-handler.ts#L5, https://github.com/nestjs/nest/blob/eedd9df6e42aa035d14c2e861f401269d26d4d4f/sample/27-scheduling/src/tasks/tasks.service.ts#L6

Also I have found that passing `logger: false` to the `NestFactory.create` can swallow some important initialization errors which are logged before `app.useLogger`. This was really not obvious from the docs.

## What is the new behavior?

I have created a separate section in the docs, explaining how the logger instantiation works when custom logger is used:
```typescript
import { Logger, Injectable } from '@nestjs/common';

@Injectable()
class MyService {
  logger = new Logger(MyService.name);
  
  doSomething() {
    // Actually uses MyLogger
    this.logger.log('Doing something...');
  }
}
```

```typescript
const app = await NestFactory.create(ApplicationModule);
app.useLogger(app.get(MyLogger));
await app.listen(3000);
``` 

I have repurposed existing section about injecting custom logger with transient scope into some kind of an advanced example where user may want to use custom logging methods which are not provided by the `LoggerService` interface.

I have also added explicit warning about using `logger: false`, maybe it will save someone hours of debugging.

```
[ ] Yes
[x] No
```

I have preserved the original section name so links should not be broken.

## Other information

Since I am not a native english speaker, there may be some stylistic errors in my descriptions. Please tell me about them so I can improve.

In short, I want to promote the best practice of using `new Logger` since it fells more consistent than just injecting your logger into your own services. For example, if all library authors use this pattern, it would be easy for me as an app developer to customize logger behavior consistently across all application.

### DIFF
--- a/content/techniques/logger.md
+++ b/content/techniques/logger.md
@@ -189,8 +189,6 @@ import { Injectable, Scope, Logger } from '@nestjs/common';
 
 @Injectable({ scope: Scope.TRANSIENT })
 export class MyLogger extends Logger {
-  // Implement custom methods or extend default methods here,
-  // for example:
   customLog() {
     this.log('Please feed the cat!');
   }
--- a/content/techniques/logger.md
+++ b/content/techniques/logger.md
@@ -165,6 +165,7 @@ class MyService {
 ```
 
 In the default logger `context` is printed in the square brackets, like `NestFactory` in the example below:
+
 ```bash
 [Nest] 19096   - 12/08/2019, 7:12:59 AM   [NestFactory] Starting Nest application...
 ```
--- a/content/techniques/logger.md
+++ b/content/techniques/logger.md
@@ -169,11 +169,11 @@ In the default logger implementation, `context` is printed in the square bracket
 [Nest] 19096   - 12/08/2019, 7:12:59 AM   [NestFactory] Starting Nest application...
 ```
 
-If we supply a custom logger via `app.useLogger`, it will actually be used by Nest automatically.
-That means that our code remains implementation agnostic, while we can easily substitute the default logger for our custom one by calling `app.useLogger`.
-That way if we follow the steps from previous section and call `app.useLogger(app.get(MyLogger))`, the following calls to `this.logger.log` from `MyService` would result in calls to method `log` from `MyLogger` instance.
+If we supply a custom logger via `app.useLogger()`, it will actually be used by Nest internally. That means that our code remains implementation agnostic, while we can easily substitute the default logger for our custom one by calling `app.useLogger()`.
 
-This should be suitable for most cases. But if you need more customization for the Logger (like adding and calling custom methods), please follow to the next section. 
+That way if we follow the steps from the previous section and call `app.useLogger(app.get(MyLogger))`, the following calls to `this.logger.log()` from `MyService` would result in calls to method `log` from `MyLogger` instance.
+
+This should be suitable for most cases. But if you need more customization (like adding and calling custom methods), move to the next section. 
 
 #### Injecting a custom logger
 
--- a/content/techniques/logger.md
+++ b/content/techniques/logger.md
@@ -148,8 +148,7 @@ You can also inject this `MyLogger` provider in your feature classes, thus ensur
 
 We can combine several of the techniques above to provide consistent behavior and formatting across both Nest system logging and our own application event/message logging.
 
-The best practice is to instantiate `Logger` class from `@nestjs/common` in each of our services. 
-We can supply our service name as the `context` argument in the `Logger` constructor, like so:
+A good practice is to instantiate `Logger` class from `@nestjs/common` in each of our services. We can supply our service name as the `context` argument in the `Logger` constructor, like so:
 
 ```typescript
 import { Logger, Injectable } from '@nestjs/common';
--- a/content/techniques/logger.md
+++ b/content/techniques/logger.md
@@ -177,7 +177,6 @@ This should be suitable for most cases. But if you need more customization (like
 
 #### Injecting a custom logger
 
-
 To start, extend the built-in logger with code like the following. We supply the `scope` option as configuration metadata for the `Logger` class, specifying a [transient](/fundamentals/injection-scopes) scope, to ensure that we'll have a unique instance of the `MyLogger` in each feature module. In this example, we do not extend the individual `Logger` methods (like `log()`, `warn()`, etc.), though you may choose to do so.
 
 ```typescript
--- a/content/techniques/logger.md
+++ b/content/techniques/logger.md
@@ -155,7 +155,7 @@ import { Logger, Injectable } from '@nestjs/common';
 
 @Injectable()
 class MyService {
-  logger = new Logger(MyService.name);
+  private readonly logger = new Logger(MyService.name);
   
   doSomething() {
     this.logger.log('Doing something...');
--- a/content/techniques/logger.md
+++ b/content/techniques/logger.md
@@ -177,10 +177,6 @@ This should be suitable for most cases. But if you need more customization (like
 
 #### Injecting a custom logger
 
-Sometimes you may want to add custom methods to your logger class.
-In order to use these methods you'll have to inject a custom logger to your services.
-We can use a [transient](/fundamentals/injection-scopes) scope for the logger so that each one of our services has its own custom context.
-If we maintain the `LoggerService` interface in our custom logger, we can still set our logger globally with `app.useLogger`.
 
 To start, extend the built-in logger with code like the following. We supply the `scope` option as configuration metadata for the `Logger` class, specifying a [transient](/fundamentals/injection-scopes) scope, to ensure that we'll have a unique instance of the `MyLogger` in each feature module. In this example, we do not extend the individual `Logger` methods (like `log()`, `warn()`, etc.), though you may choose to do so.
 
--- a/content/techniques/logger.md
+++ b/content/techniques/logger.md
@@ -246,8 +246,7 @@ app.useLogger(new MyLogger());
 await app.listen(3000);
 ```
 
-Be mindful that if you supply `logger: false` to `NestFactory.create`, nothing will be logged until you call `useLogger`, so you may miss some important initialization errors.
-If you don't mind that some of your initial messages will be logged with the default logger, you can just omit the `logger: false` option.
+Be mindful that if you supply `logger: false` to `NestFactory.create`, nothing will be logged until you call `useLogger`, so you may miss some important initialization errors. If you don't mind that some of your initial messages will be logged with the default logger, you can just omit the `logger: false` option.
 
 #### Use external logger
 
--- a/content/techniques/logger.md
+++ b/content/techniques/logger.md
@@ -163,7 +163,7 @@ class MyService {
 }
 ```
 
-In the default logger `context` is printed in the square brackets, like `NestFactory` in the example below:
+In the default logger implementation, `context` is printed in the square brackets, like `NestFactory` in the example below:
 
 ```bash
 [Nest] 19096   - 12/08/2019, 7:12:59 AM   [NestFactory] Starting Nest application...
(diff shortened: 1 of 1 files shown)


content/techniques/logger.md
```diff
@@ -133,30 +133,68 @@ await app.listen(3000);
...
+The best practice is to instantiate `Logger` class from `@nestjs/common` in each of our services. 
+We can supply our service name as the `context` argument in the `Logger` constructor, like so:
 
-1. We extend the built-in logger and customize the `context` portion of the log message (e.g., the phrase `NestFactory` in square brackets in the log line shown below).
+```typescript
+import { Logger, Injectable } from '@nestjs/common';
+
+@Injectable()
+class MyService {
+  logger = new Logger(MyService.name);
```
jmcdo29, line 159: It would be better to show this in the `constructor` to better follow dependency injection principles.

content/techniques/logger.md
```diff
@@ -133,30 +133,68 @@ await app.listen(3000);
...
+  
+  doSomething() {
+    this.logger.log('Doing something...');
+  }
+}
+```
 
+In the default logger `context` is printed in the square brackets, like `NestFactory` in the example below:
 ```bash
```
jmcdo29, line 168: Add a newline to separate the code block from the regular text.

content/techniques/logger.md
mayorandrew: Do you mean like this?
```
constructor() {
  this.logger = new Logger(MyService.name);
}
```

From my standpoint it is no different from what wrote, but through the times of my googling I only saw variant with property initializer (even in the nestjs source), so I used it here as well.

jmcdo29: Rather something like
```ts
class MyService {
  constructor(private readonly logger: Logger) {
    this.logger.setContext(MyService.name);
  }
}
```

content/techniques/logger.md
mayorandrew: Done, thanks for noticing

content/techniques/logger.md
mayorandrew: I had some issues with that with `nestjs-pino` because it has its own provider for `Logger`, so this code actually injects `Logger` from `nestjs-pino` instead of `Logger` from `@nestjs/common`. And the former does not have `setContext` function. So it produces a runtime error. Maybe this is a mistake with the library, but still `new Logger` seems more safe.

Also, `setContext` mutates `Logger` instance and `Logger` seems to have singleton scope (https://github.com/nestjs/nest/blob/master/packages/common/services/logger.service.ts#L19), so I am not sure this would work.

jmcdo29: Ah, in the case of singleton scopes, you'd be right about that. In the case of `nestjs-pino` I feel that's on the library. You could always get around that with `@Inject()`, but that's not what this PR is about ;). All right, LGTM.

content/techniques/logger.md
```diff
@@ -133,30 +133,69 @@ await app.listen(3000);
...
+
+You can also inject this `MyLogger` provider in your feature classes, thus ensuring consistent logging behavior across both Nest system logging and application logging. See <a href="techniques/logger#using-the-logger-for-application-logging">Using the logger for application logging</a> and <a href="techniques/logger#injecting-a-custom-logger">Injecting a custom logger</a> below for more information.
 
 #### Using the logger for application logging
 
-We can combine several of the techniques above to provide consistent behavior and formatting across both Nest system logging and our own application event/message logging. In this section, we'll achieve this with the following steps:
+We can combine several of the techniques above to provide consistent behavior and formatting across both Nest system logging and our own application event/message logging.
+
+The best practice is to instantiate `Logger` class from `@nestjs/common` in each of our services. 
+We can supply our service name as the `context` argument in the `Logger` constructor, like so:
```
kamilmysliwiec, line 152:
```suggestion
A good practice is to instantiate `Logger` class from `@nestjs/common` in each of our services. We can supply our service name as the `context` argument in the `Logger` constructor, like so:
```

content/techniques/logger.md
```diff
@@ -133,30 +133,68 @@ await app.listen(3000);
...
+
+A good practice is to instantiate `Logger` class from `@nestjs/common` in each of our services. We can supply our service name as the `context` argument in the `Logger` constructor, like so:
 
-1. We extend the built-in logger and customize the `context` portion of the log message (e.g., the phrase `NestFactory` in square brackets in the log line shown below).
+```typescript
+import { Logger, Injectable } from '@nestjs/common';
+
+@Injectable()
+class MyService {
+  logger = new Logger(MyService.name);
```
kamilmysliwiec, line 158:
```suggestion
  private readonly logger = new Logger(MyService.name);
```

content/techniques/logger.md
```diff
@@ -133,30 +133,68 @@ await app.listen(3000);
...
+class MyService {
+  private readonly logger = new Logger(MyService.name);
+  
+  doSomething() {
+    this.logger.log('Doing something...');
+  }
+}
+```
+
+In the default logger `context` is printed in the square brackets, like `NestFactory` in the example below:
```
kamilmysliwiec, line 166:
```suggestion
In the default logger implementation, `context` is printed in the square brackets, like `NestFactory` in the example below:
```

content/techniques/logger.md
```diff
@@ -133,30 +133,68 @@ await app.listen(3000);
...
 [Nest] 19096   - 12/08/2019, 7:12:59 AM   [NestFactory] Starting Nest application...
 ```
 
-2. We inject a [transient](/fundamentals/injection-scopes) instance of the `Logger` into our feature modules so that each one has its own custom context.
-3. We supply this extended logger for Nest to use for system logging.
+If we supply a custom logger via `app.useLogger`, it will actually be used by Nest automatically.
+That means that our code remains implementation agnostic, while we can easily substitute the default logger for our custom one by calling `app.useLogger`.
+That way if we follow the steps from previous section and call `app.useLogger(app.get(MyLogger))`, the following calls to `this.logger.log` from `MyService` would result in calls to method `log` from `MyLogger` instance.
+
+This should be suitable for most cases. But if you need more customization for the Logger (like adding and calling custom methods), please follow to the next section.
```
kamilmysliwiec, line 176:
```suggestion
If we supply a custom logger via `app.useLogger()`, it will actually be used by Nest internally. That means that our code remains implementation agnostic, while we can easily substitute the default logger for our custom one by calling `app.useLogger()`.

That way if we follow the steps from the previous section and call `app.useLogger(app.get(MyLogger))`, the following calls to `this.logger.log()` from `MyService` would result in calls to method `log` from `MyLogger` instance.

This should be suitable for most cases. But if you need more customization (like adding and calling custom methods), move to the next section. 
```

content/techniques/logger.md
```diff
@@ -203,6 +246,9 @@ app.useLogger(new MyLogger());
 await app.listen(3000);
 ```
 
+Be mindful that if you supply `logger: false` to `NestFactory.create`, nothing will be logged until you call `useLogger`, so you may miss some important initialization errors.
+If you don't mind that some of your initial messages will be logged with the default logger, you can just omit the `logger: false` option.
```
kamilmysliwiec, line 250:
```suggestion
Be mindful that if you supply `logger: false` to `NestFactory.create`, nothing will be logged until you call `useLogger`, so you may miss some important initialization errors. If you don't mind that some of your initial messages will be logged with the default logger, you can just omit the `logger: false` option.
```

content/techniques/logger.md
```diff
@@ -133,30 +133,68 @@ await app.listen(3000);
...
+To start, extend the built-in logger with code like the following. We supply the `scope` option as configuration metadata for the `Logger` class, specifying a [transient](/fundamentals/injection-scopes) scope, to ensure that we'll have a unique instance of the `MyLogger` in each feature module. In this example, we do not extend the individual `Logger` methods (like `log()`, `warn()`, etc.), though you may choose to do so.
 
 ```typescript
 import { Injectable, Scope, Logger } from '@nestjs/common';
 
 @Injectable({ scope: Scope.TRANSIENT })
-export class MyLogger extends Logger {}
+export class MyLogger extends Logger {
+  // Implement custom methods or extend default methods here,
+  // for example:
```
kamilmysliwiec, line 193:
```suggestion
```

content/techniques/logger.md
```diff
@@ -133,30 +133,66 @@ await app.listen(3000);
...
+
+This should be suitable for most cases. But if you need more customization (like adding and calling custom methods), move to the next section. 
+
+#### Injecting a custom logger
 
-To start, extend the built-in logger with code like the following. We supply the `scope` option as configuration metadata for the `Logger` class, specifying a [transient](/fundamentals/injection-scopes) scope, to ensure that we'll have a unique instance of the `Logger` in each feature module. In this example, we do not extend the individual `Logger` methods (like `log()`, `warn()`, etc.), though you may choose to do so.
+Sometimes you may want to add custom methods to your logger class.
+In order to use these methods you'll have to inject a custom logger to your services.
+We can use a [transient](/fundamentals/injection-scopes) scope for the logger so that each one of our services has its own custom context.
+If we maintain the `LoggerService` interface in our custom logger, we can still set our logger globally with `app.useLogger`.
```
kamilmysliwiec, line 183:
```suggestion
```

content/techniques/logger.md
```diff
@@ -133,30 +133,62 @@ await app.listen(3000);
...
+If we supply a custom logger via `app.useLogger()`, it will actually be used by Nest internally. That means that our code remains implementation agnostic, while we can easily substitute the default logger for our custom one by calling `app.useLogger()`.
+
+That way if we follow the steps from the previous section and call `app.useLogger(app.get(MyLogger))`, the following calls to `this.logger.log()` from `MyService` would result in calls to method `log` from `MyLogger` instance.
+
+This should be suitable for most cases. But if you need more customization (like adding and calling custom methods), move to the next section. 
+
+#### Injecting a custom logger
 
-To start, extend the built-in logger with code like the following. We supply the `scope` option as configuration metadata for the `Logger` class, specifying a [transient](/fundamentals/injection-scopes) scope, to ensure that we'll have a unique instance of the `Logger` in each feature module. In this example, we do not extend the individual `Logger` methods (like `log()`, `warn()`, etc.), though you may choose to do so.
+
```
kamilmysliwiec, line 180:
```suggestion
```

PR #1449, opened 2020-09-03
fix Unknown authentication strategy "local" error

bahmanbs:
additional description that helps to fix Unknown authentication strategy "local" error

### DIFF
--- a/content/techniques/authentication.md
+++ b/content/techniques/authentication.md
@@ -352,7 +352,9 @@ export class AppController {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


ruslanguns:
I believe that is quite obvious to understand it, if you thread with Modules inside your NestJS architecture you should import your module in the parent module. 

Also!, if you follow the docs instructions, the AuthModule has been generated with the NestJS CLI so the schematics already adds the importation correctly, therefore the content is very clear. However, if you still want to clarify more, let it be considered.

PR #1482, opened 2020-09-29
mention another way to consume OpenAPIObject document

jokester:
## PR Checklist

## PR Type

[x] Document

People unfamiliar with OpenAPI may fail to realize other ways to use the `OpenAPIObject` object (at least in my own case).

## What is the new behavior?

It is clearer that one could save the OpenAPI document as a file, rather than serving it over HTTP with `SwaggerModule.setup`.

[ ] Yes
[x] No

## Other information

The use I described in this patch worked in my project but I'm not very confident though. Can somebody confirm this usage?


jmcdo29 (approved):

kamilmysliwiec:
Thanks!

PR #1498, opened 2020-10-05
fix(a11y): make the menu button accessible

dianjuar:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] Other... Please describe:
```

Issue Number: N/A

The menu button is not accessible by the keyboard.

old

## What is the new behavior?

The menu now is accessible by keyboard

new

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/src/app/homepage/header/header.component.html
+++ b/src/app/homepage/header/header.component.html
@@ -1,6 +1,7 @@
 <header>
-  <div
-    class="icon-wrapper"
+  <button
+    title="Docs Menu"
+    class="icon-wrapper no-button"
     (click)="toggle.emit()"
     [class.opened]="isSidebarOpened"
   >
@@ -38,7 +39,7 @@
         />
       </svg>
     </div>
-  </div>
+  </button>
   <ng-content></ng-content>
   <app-social-wrapper></app-social-wrapper>
   <div class="menu-wrapper">
--- a/src/app/homepage/header/header.component.scss
+++ b/src/app/homepage/header/header.component.scss
@@ -25,14 +25,9 @@ header {
   @extend .transition;
   height: 100%;
   width: 80px;
-  top: 0;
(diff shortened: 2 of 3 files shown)


kamilmysliwiec:
Amazing contribution @dianjuar, thank you!

One note: could we remove the outline on **click**?

image

dianjuar:
> Amazing contribution @dianjuar, thank you!
> 
> One note: could we remove the outline on **click**?
> 
> image

You are right, I'll do it

kamilmysliwiec:
@dianjuar any chance you can look into this in the next few weeks 💪 ?

dianjuar:
Yes, for sure.
I don't want to let this PR die.

dianjuar:
Hi @kamilmysliwiec 
It's a chromes behavior, that's the way chrome manages the focus of a button without background
```css
background: none;
```
Prove that buttons without background always mark the outline when is focused

chrome-prove

If you look the page with a different browser if has the expected behavior

**Firefox**

firefox

dianjuar:
It can be done but it's not [that straight forward](https://www.kizu.ru/keyboard-only-focus/).
I think it worth the effort.

kamilmysliwiec:
@dianjuar it seems that Angular docs (https://angular.io/docs) have a different CSS applied on focus:

image

vs

image

Could we do something similar instead of having this blue outline?

dianjuar:
In the Angular official web page, they have that behavior because they are using AngularMaterial, it normalizes all the focus look and feel to look very similar across all the browsers using javascript and custom CSS classes to achieve that behavior.

I would have to implement CSS classes and custom directives to achieve that behavior or use AngularMaterial buttons to achieve that behavior.

angular material

PR #1593, opened 2020-12-09
docs(sql): how to export custom repositories

alefcastelo:
## PR Checklist

## PR Type
An example of how to export custom repositories to be used as a dependency on other modules.

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[x] Docs
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[x] No
```


kamilmysliwiec:
So this actually isn't the best idea in all cases. Sometimes, it's better to simply re-register the repository **again** in another module instead of importing the whole module as this may introduce complex circular dependencies. This instruction (from this PR) could confuse devs :( Thanks for your contribution anyway!

Issue #1326, opened 2020-06-28
Add type to the UploadedFile argument


incompletude:
## Is your feature request related to a problem? Please describe.

Took me a while to find the proper type to a `file` from `@UploadedFile()` decorator.

## Describe the solution you'd like

Add the following to the docs:

`npm i -D @types/multer` 

```
  @Post()
  @UseInterceptors(FileInterceptor("file"))
  uploadFile(@UploadedFile() file: Express.Multer.File): void {
    console.log(file)
  }
```

kamilmysliwiec:
PRs are more than welcome!

incompletude:
I will send a PR. Thx.

kamilmysliwiec:
Let's track this here https://github.com/nestjs/docs.nestjs.com/pull/1604

PR #1604, opened 2020-12-15
docs(file-upload): added types for UploadedFile argument

Tony133:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[x] Docs
[ ] Other... Please describe:
```

Issue Number: N/A

## What is the new behavior?
N/A

```
[ ] Yes
[x] No
```

## Other information
This PR, added types for UploadedFile argument

### DIFF
--- a/content/techniques/file-upload.md
+++ b/content/techniques/file-upload.md
@@ -4,6 +4,14 @@ To handle file uploading, Nest provides a built-in module based on the [multer](
 
 > warning **Warning** Multer cannot process data which is not in the supported multipart format (`multipart/form-data`). Also, note that this package is not compatible with the `FastifyAdapter`.
 
+#### Use with Express (default)
+
+To work properly with Typescript, its types must be installed:
+
+```shell
+$ npm i -D @types/multer
+```
+
 #### Basic example
 
 To upload a single file, simply tie the `FileInterceptor()` interceptor to the route handler and extract `file` from the `request` using the `@UploadedFile()` decorator.
@@ -12,7 +20,7 @@ To upload a single file, simply tie the `FileInterceptor()` interceptor to the r
 @@filename()
 @Post('upload')
 @UseInterceptors(FileInterceptor('file'))
-uploadFile(@UploadedFile() file) {
+uploadFile(@UploadedFile() file: Express.Multer.File) {
   console.log(file);
 }
 @@switch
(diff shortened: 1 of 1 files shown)


Tony133:
reference issue: https://github.com/nestjs/docs.nestjs.com/issues/1326

jmcdo29 (approved):

content/techniques/file-upload.md
```diff
@@ -4,6 +4,14 @@ To handle file uploading, Nest provides a built-in module based on the [multer](
 
 > warning **Warning** Multer cannot process data which is not in the supported multipart format (`multipart/form-data`). Also, note that this package is not compatible with the `FastifyAdapter`.
 
+#### Use with Express (default)
+
+To work properly with Typescript, its types must be installed:
```
johannesschobel, line 9:
```suggestion
First install the types of the required package to properly work with Typescript:
```

johannesschobel (changes requested):

content/techniques/file-upload.md
jmcdo29: I think either of these is fine here. Maybe we could use something like

```suggestion
If you want to use the `Express.Multer.File` type for `@UploadedFile()` files, it is necessary to install the types:
```

to show that this is a completely optional step and only necessary if the dev wants to use this type.

Tony133: @jmcdo29  i think if we really have to make a change, I'd say use yours as a change, ok?

content/techniques/file-upload.md
```diff
@@ -4,6 +4,14 @@ To handle file uploading, Nest provides a built-in module based on the [multer](
...
 > warning **Warning** Multer cannot process data which is not in the supported multipart format (`multipart/form-data`). Also, note that this package is not compatible with the `FastifyAdapter`.
 
+#### Use with Express (default)
+
+If you want to use the `Express.Multer.File` type for `@UploadedFile()` files, it is necessary to install the types:
+
+```shell
+$ npm i -D @types/multer
+```
+
```
kamilmysliwiec, line 14: Instead of adding this as a heading, I think we should rather add a blockquote (hint) + the heading title is somewhat misleading, especially the "default" part given that this doesn't work with Fastify at all (and after reading this heading, I would have assumed there's an alternative way covered)

Tony133: For me it is indifferent, @jmcdo29  what do you think?

PR #1623, opened 2020-12-29
docs(microservices): added hint for mqtt

Tony133:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] Docs
[x] Other... Please describe:
```

Issue Number: #1120

## What is the new behavior?
N/A

```
[ ] Yes
[x] No
```

## Other information
This PR, added hint for mqtt, in case you are using a docker environment with MQTT

### DIFF
--- a/content/microservices/mqtt.md
+++ b/content/microservices/mqtt.md
@@ -14,7 +14,7 @@ $ npm i --save mqtt
 
 To use the MQTT transporter, pass the following options object to the `createMicroservice()` method:
 
-> info **Hint** If you were to use MQTT in a docker environment I recommend using `hostname` as a parameter options.
+> info **Hint** In development environments where you need to specify the port number, use `hostname` instead of `host`
 
 ```typescript
 @@filename(main)
--- a/content/microservices/mqtt.md
+++ b/content/microservices/mqtt.md
@@ -14,6 +14,8 @@ $ npm i --save mqtt
 
 To use the MQTT transporter, pass the following options object to the `createMicroservice()` method:
 
+> info **Hint** If you were to use MQTT in a docker environment I recommend using `hostname` as a parameter options.
+
 ```typescript
 @@filename(main)
 const app = await NestFactory.createMicroservice<MicroserviceOptions>(ApplicationModule, {


content/microservices/mqtt.md
```diff
@@ -14,6 +14,8 @@ $ npm i --save mqtt
 
 To use the MQTT transporter, pass the following options object to the `createMicroservice()` method:
 
+> info **Hint** If you were to use MQTT in a docker environment I recommend using `hostname` as a parameter options.
```
jmcdo29, line 17: Why would we want to add this to the docs? This feels more like a comment that would belong in a blog post

Tony133: It might help partly as there is a little confusion between host and hostname, maybe I can rewrite the suggestion differently

Tony133: "In a docker environment it is recommended to use hostname instead of host"
Maybe this is how it could go?

jmcdo29: Possibly, but it still seems out of place, as there's not really anything else in the docs that mentions docker. This feels like it should be one of those things where understanding docker and what the properties means is more valuable than having documentation tell you what to set. Especially when with Nest it's just _using_ RabbitMQ and not setting it up directly.

Tony133: "In development environments where it is necessary to specify the port instead of host use hostname"

kamilmysliwiec:
I do agree with @jmcdo29 here, if we started adding such hints to the documentation, we'd eventually have hundreds of library-specific, additional tips & tricks making the docs harder to follow.

Tony133:
@kamilmysliwiec then this issue #1120 can be closed

dcs3spp:
@kamilmysliwiec @Tony133 One possiblity for issues such as [#1120](https://github.com/nestjs/docs.nestjs.com/issues/1120) could be adding it to FAQ of the docs. Alternatively introducing Troubleshooting section for issues such as this?? This might save time and reduce the learning curve for developers / businesses who may be considering using NestJS.

jmcdo29:
@dcs3spp as Kamil said, if this is the case we'd need to do it for every integration that is mentioned in the docs. Setting up databases through Docker (MySQL, MSSQL, PostgreSQL, Mongo), and every microservice Nest supports (meaning Kafka, Redis, RabbitMQ, Redis, and MQTT). All of these should have their own documentation as they are all their own technology. Usually you can find this kind of information on DockerHub, or relevant tutorials around the web.

Tony133:
I agree with what @jmcdo29 and @kamilmysliwiec said.

PR #1628, opened 2020-12-31
fix(security/helmet): fix import statement

stingalleman:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[X] Docs
[ ] Other... Please describe:
```

Issue Number: N/A

## What is the new behavior?

`import * as helmet from 'helmet'` errors, it should be `import helmet from 'helmet'`. 

Error:
```
Type originates at this import. A namespace-style import cannot be called or constructed, and will cause a failure at runtime. Consider using a default import or import require here instead.
```

```
[ ] Yes
[X] No
```

### DIFF
--- a/content/security/helmet.md
+++ b/content/security/helmet.md
@@ -15,7 +15,7 @@ $ npm i --save helmet
 Once the installation is complete, apply it as a global middleware.
 
 ```typescript
-import * as helmet from 'helmet';
+import helmet from 'helmet';
 // somewhere in your initialization file
 app.use(helmet());
 ```
--- a/content/security/helmet.md
+++ b/content/security/helmet.md
@@ -20,7 +20,7 @@ import * as helmet from 'helmet';
 app.use(helmet());
 ```
 
-> info **Hint** If you are getting a `This expression is not callable` error while trying to import `Helmet`, you very likely have the `allowSyntheticDefaultImports` and `esModuleInterop` options set to `true` in your project's `tsconfig.json` file. If that's the case, change the import statement to: `import helmet from 'helmet'` instead.
+> info **Hint** If you are getting the `This expression is not callable` error while trying to import `Helmet`, you very likely have the `allowSyntheticDefaultImports` and `esModuleInterop` options set to `true` in your project's `tsconfig.json` file. If that's the case, change the import statement to: `import helmet from 'helmet'` instead.
 
 #### Use with Fastify
 
--- a/content/security/helmet.md
+++ b/content/security/helmet.md
@@ -15,12 +15,12 @@ $ npm i --save helmet
 Once the installation is complete, apply it as a global middleware.
 
 ```typescript
-import helmet from 'helmet';
+import * as helmet from 'helmet';
 // somewhere in your initialization file
 app.use(helmet());
 ```
 
-> info **Hint** If you are getting a error (`TypeError: helmet_1.default is not a function`) while trying to import Helmet, you will need to set `allowSyntheticDefaultImports` and `esModuleInterop` to `true` in your project's `tsconfig.json`. Another solution is to import `Helmet` using `import * as helmet from 'helmet'`.
+> info **Hint** If you are getting a error (`This expression is not callable`) while trying to import Helmet, you probably have `allowSyntheticDefaultImports` and `esModuleInterop` set to `true` in your project's `tsconfig.json`. If this is the case, you will need to use `import helmet from 'helmet'`, instead of the example.
 
 #### Use with Fastify
 
--- a/content/security/helmet.md
+++ b/content/security/helmet.md
@@ -20,7 +20,7 @@ import * as helmet from 'helmet';
 app.use(helmet());
 ```
 
-> info **Hint** If you are getting a error (`This expression is not callable`) while trying to import Helmet, you probably have `allowSyntheticDefaultImports` and `esModuleInterop` set to `true` in your project's `tsconfig.json`. If this is the case, you will need to use `import helmet from 'helmet'`, instead of the example.
+> info **Hint** If you are getting a `This expression is not callable` error while trying to import `Helmet`, you very likely have the `allowSyntheticDefaultImports` and `esModuleInterop` options set to `true` in your project's `tsconfig.json` file. If that's the case, change the import statement to: `import helmet from 'helmet'` instead.
 
 #### Use with Fastify
 
--- a/content/security/helmet.md
+++ b/content/security/helmet.md
@@ -20,6 +20,8 @@ import helmet from 'helmet';
 app.use(helmet());
 ```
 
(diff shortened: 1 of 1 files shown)


keesvv (approved):

content/security/helmet.md
```diff
@@ -15,7 +15,7 @@ $ npm i --save helmet
 Once the installation is complete, apply it as a global middleware.
 
 ```typescript
-import * as helmet from 'helmet';
+import helmet from 'helmet';
```
jmcdo29, line 18: Why the change? Using the newest version of `helmet` I'm able to use `import * as helmet from 'helmet'` without any warning or errors. Node version 14.5.0

stingalleman: Also on the latest version of `helmet`, node version 14.15.3 and the latest TypeScript SDK. I'm getting a compiler error: 
image

jmcdo29: Still not getting the same message to show up. 

Updated node to 14.15.3
Updated Typescript to 4.1.3
Updated helmet at 4.3.1
Using `import * as helmet from 'helmet'` without an issue. 

Same functionality with both `yarn` and `npm`

stingalleman: I don't know why it's working on your side but errors on mine, but I'm sure this is a good change. `Helmet` exports a default in `index.ts` (see [Helmet index.ts](https://github.com/helmetjs/helmet/blob/main/index.ts) and this 
image, it clearly shows that it's a default type)

jmcdo29: [Line 201](https://github.com/helmetjs/helmet/blob/main/index.ts#L201) `export = helmet`. There's no `default` there, and there's no default in the `d.ts`

jmcdo29: You _can_ do a `default` import (`import helmet from 'helmet'`) but it isn't neccessary

stingalleman: Yes, that's correct. The [TypeScript docs](https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require) indicate that `export =` (in `helmet` index.ts) requires you to do `import helmet = require('helmet')`. That's the same as `import helmet from 'helmet'. Also see this [SO answer](https://stackoverflow.com/a/35706271), that confirms my claims.

stingalleman: > You _can_ do a `default` import (`import helmet from 'helmet'`) but it isn't neccessary

I missed this one. Yes, you can, but the SO answer linked above says this is not recommended.

> Attempting to use `import * as express` and then invoking `express()` is always illegal according to the ES6 spec. In some runtime+transpilation environments this might happen to work anyway, but it might break at any point in the future without warning, which will make you sad.

jmcdo29: Interesting. I wonder if some of this comes down to what `target` you're configured for

jmcdo29 (approved):
After the great discussion, I think should be a fine change to make. It's not that it's a default export, but an ES6 quirk/oddity. Maybe `import helmet = require('helmet')` would be better, but overall, I think this should be fine.

stingalleman:
Just out of curiosity @jmcdo29, can you share the `tsconfig` you were using?

Mine: https://github.com/stingalleman/boilerplate.ts/blob/main/tsconfig.json

jmcdo29:
@stingalleman just Nest's default tsconfig

```
{
  "compilerOptions": {
    "module": "commonjs",
    "declaration": true,
    "removeComments": true,
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true,
    "allowSyntheticDefaultImports": true,
    "allowJs": true,
    "target": "es2017",
    "sourceMap": true,
    "outDir": "./dist",
    "baseUrl": "./",
    "incremental": true
  }
}
```

stingalleman:
@jmcdo29 I guess it has something to do with this, maybe the target. I was just curious why this was happening.

kamilmysliwiec:
With `import helmet from 'helmet'`, the default Nest boilerplate (`typescript-starter` or this one generated by `nest new`) will crash with the following error:

```
(node:9173) UnhandledPromiseRejectionWarning: TypeError: helmet_1.default is not a function
```
To make it work properly, you must set the `esModuleInterop` to `true` in the `tsconfig.json` (it's disabled by default). 

On the other hand, when you have the `esModuleInterop` enabled, `import * as helmet` syntax won't work ("This expression is not callable" error). 

Would you like to update this PR with a hint (blockquote) that a different import statement must be used depending on the `esModuleInterop` configuration? @stingalleman

stingalleman:
@kamilmysliwiec Pushed!

jmcdo29 (commented):
I feel this should be the other way around. In the docs we keep the `import * as helmet from 'helmet'` and then mention the error you _might_ get with `esModuleInterop`, as a `nest new` project works with `import * as helmet from 'helmet'` and not `import helmet from 'helmet'`.

stingalleman:
I agree, that's better.

stingalleman:
Again, @jmcdo29, @kamilmysliwiec , I pushed

content/security/helmet.md
```diff
@@ -20,6 +20,8 @@ import * as helmet from 'helmet';
 app.use(helmet());
 ```
 
+> info **Hint** If you are getting a error (`This expression is not callable`) while trying to import Helmet, you probably have `allowSyntheticDefaultImports` and `esModuleInterop` set to `true` in your project's `tsconfig.json`. If this is the case, you will need to use `import helmet from 'helmet'`, instead of the example.
```
kamilmysliwiec, line 23:
```suggestion
> info **Hint** If you are getting a `This expression is not callable` error while trying to import `Helmet`, you very likely have the `allowSyntheticDefaultImports` and `esModuleInterop` options set to `true` in your project's `tsconfig.json` file. If that's the case, change the import statement to: `import helmet from 'helmet'` instead.

File: content/security/helmet.md [hunk 3]
```diff
@@ -20,6 +20,8 @@ import * as helmet from 'helmet';
 app.use(helmet());
 ```
 
+> info **Hint** If you are getting a `This expression is not callable` error while trying to import `Helmet`, you very likely have the `allowSyntheticDefaultImports` and `esModuleInterop` options set to `true` in your project's `tsconfig.json` file. If that's the case, change the import statement to: `import helmet from 'helmet'` instead.
```
kamilmysliwiec, line 23:
```suggestion
> info **Hint** If you are getting the `This expression is not callable` error while trying to import `Helmet`, you very likely have the `allowSyntheticDefaultImports` and `esModuleInterop` options set to `true` in your project's `tsconfig.json` file. If that's the case, change the import statement to: `import helmet from 'helmet'` instead.
```

kamilmysliwiec:
Thanks!

PR #1671, opened 2021-01-22
Add a warning about defining hooks

msrumon:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[x] Docs
[ ] Other... Please describe:
```

There is no information about the fact that defining hooks using `MongooseModule.forFeatureAsync()` doesn't need to separately declare the schema via `MongooseModule.forFeature()` and that it won't work if it's done.

Issue Number: N/A

## What is the new behavior?
A warning message has been added to state that information.

```
[ ] Yes
[x] No
```

### DIFF
--- a/content/techniques/mongo.md
+++ b/content/techniques/mongo.md
@@ -275,6 +275,8 @@ Middleware (also called pre and post hooks) are functions which are passed contr
 export class AppModule {}
 ```
 
+> warning **NOTICE** Since you're returning the schema itself from the factory function, it's not required to have `forFeature()` separately and declare the schema there. In fact, the hook won't be triggered if you do so and it may create side-effects in some situations.
+
 Like other [factory providers](https://docs.nestjs.com/fundamentals/custom-providers#factory-providers-usefactory), our factory function can be `async` and can inject dependencies through `inject`.
 
 ```typescript


kamilmysliwiec:
I'm not sure if I'm following. `forFeature` and `forFeatureAsync` should never be used for the same schema twice, and this applies to pretty much every module in NestJS ecosystem when there's a way for both static and asynchronous registrations. Using `forFeature` and `forFeatureAsync` at the same time (for the same target schema/entity/connection configuration/cache or config configuration) is like calling the same function two times and if this isn't clear, we should rather add a more "generic" note that these are the same methods used depending on whether you need to do something synchronously or asynchronously instead of adding a blockquote about mongoose schemas (where it's information very specific to this particular case) cc @jmcdo29 do you have any ideas?

msrumon:
Oh... I just got myself stuck here, that's why I added it here. I thought it's Mongoose-specific. Anyway, if you're right, then it should definitely be put in a more generic place. Maybe <https://github.com/nestjs/docs.nestjs.com/blob/master/content/modules.md> sounds like a good place.

msrumon:
Moreover, the doc also doesn't tell that you need to call `next()`, which is automatically passed to the cb function by Mongoose. This is a very strong reason for not to use dummy examples, rather real-world examples which can benefit lots of new-comers.

jmcdo29:
> cc @jmcdo29 do you have any ideas?

Generally speaking, you're right that `forFeature` and `forFeatureAsync shouldn't ever be called together. Just due to how the underlying registration works. It would be like calling `forRoot()` and `forRootAsync()` for the same module. There _may_ be some uses cases where the latter is justifiable (different configurations in different modules) so I _guess_ technically the former could be too, but again that would be across different modules, definitely not in the same one.

As mentioned, we could probably make a note in a more common place (like Modules or Dynamic Modules), maybe something like

> Generally speaking, it is not a good idea to import a Module using both the synchronous and asynchronous registration methods. Whether this is `forRoot` and `forRootAsync`, `register` and `registerAsync`, or `forFeature` and `forFeatureAsync`, it could end up overwriting the injection tokens and the underlying configurations for the module, resulting in unexpected functionality.

I'm sure people will find reasons to do it nonetheless, but generally speaking the above holds true.

kamilmysliwiec:
Would you like to create a PR for this @msrumon

msrumon:
@kamilmysliwiec Sure. Why not?

PR #1700, opened 2021-01-28
docs(streams): adds docs for streaming files with `StreamableFiles`

jmcdo29:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[X] Docs
[ ] Other... Please describe:
```

Issue Number: N/A
There's no documentation for a new feature being implemented ([nest#5278](https://github.com/nestjs/nest/pull/5278))

## What is the new behavior?

There's now docs showing how files can be streamed back without interfering with the interceptor's post-controller logic.

```
[ ] Yes
[X] No
```

## Other information

Looks like some of the compiled output from degni got modified. Let me know if you want me to remove that or keep it. Looks like it shouldn't modify the final output to the webpage

### DIFF
--- a/content/techniques/streaming-files.md
+++ b/content/techniques/streaming-files.md
@@ -1,11 +1,16 @@
 ### Streaming Files
 
-There won't always be a time where you're sending back JSON or string responses. There may be times where you'd like to send back a file from the server to the client. In an express application, you may use something like
+There won't always be a time where you're sending back JSON or string responses. There may be times where you'd like to send back a file from the server to the client. To do this with Nest, normally you'd do te following:
 
 ```ts
-app.get('send-file', (req, res) => {
-  createReadStream(path(process.cwd(), 'package.json')).pipe(res);
-});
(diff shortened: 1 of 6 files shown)


content/techniques/streaming-files.md
```diff
@@ -0,0 +1,41 @@
+### Streaming Files
+
+There won't always be a time where you're sending back JSON or string responses. There may be times where you'd like to send back a file from the server to the client. In an express application, you may use something like
```
kamilmysliwiec, line 3:
```suggestion
There won't always be a time where you're sending back JSON or string responses. There may be times where you'd like to send back a file from the server to the client. In an express application, you may use something like
```

Maybe instead of showing an express example here, we could present how it could be accomplished using the standard, `Response` object (saying it's pretty much the same way as in native express/fastify application). The reason is that we're slowly getting more and more users that are not even familiar with express itself

PR #1703, opened 2021-01-28
Update cors.md

dorklord23:
## PR Checklist

## PR Type
Docs

The current documentation does not address how to enable CORS for GraphQL. People might be misled to believe that it also applies to GraphQL.

Issue Number: N/A

## What is the new behavior?
Explicitly tell readers that enabling CORS for GraphQL needs to be done separately.

No

### DIFF
--- a/content/security/cors.md
+++ b/content/security/cors.md
@@ -21,9 +21,7 @@ const app = await NestFactory.create(AppModule, { cors: true });
 await app.listen(3000);
 ```
 
-#### GraphQL
-
-Above method only applies to REST endpoints. To enable CORS to GraphQL, set `cors` property to `true` or pass [CORS configuration object](https://github.com/expressjs/cors#configuration-options) as the `cors` property value when you import GraphQL module.
+Above method only applies to REST endpoints. To enable CORS in GraphQL, set `cors` property to `true` or pass [CORS configuration object](https://github.com/expressjs/cors#configuration-options) as the `cors` property value when you import GraphQL module.
 
 ```typescript
 @Module({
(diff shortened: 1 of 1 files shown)


content/security/cors.md
```diff
@@ -20,3 +20,18 @@ Alternatively, enable CORS via the `create()` method's options object. Set the `
 const app = await NestFactory.create(AppModule, { cors: true });
 await app.listen(3000);
 ```
+
+#### GraphQL
```
Tony133, line 24: I would delete the title, otherwise it should also be placed above for REST endpoints, instead without I find it better.

dorklord23: So just delete the GraphQL title?

Tony133: In my opinion untitled could also go, it links well to the title above `Getting started`.

@jmcdo29  what do you think about it?

jmcdo29: I agree here, remove the GraphQL title and let it be a single page. Should be fine. If we had more than just these two titles _may_ be useful, but with just two I think this is fine.

jmcdo29 (changes requested):

content/security/cors.md
dorklord23: I've fixed it

Tony133 (approved):

content/security/cors.md
```diff
@@ -20,3 +20,16 @@ Alternatively, enable CORS via the `create()` method's options object. Set the `
...
+
+```typescript
+@Module({
+  imports: [
+    GraphQLModule.forRoot({
+      cors: {
+        origin: 'http://localhost:3000',
+        credentials: true,
+      },
+    }),
```
kamilmysliwiec, line 34:
```suggestion
GraphQLModule.forRoot({
  cors: {
    origin: 'http://localhost:3000',
    credentials: true,
  },
}),
```

kamilmysliwiec:
Thanks!

PR #1785, opened 2021-02-19
mapped-types imports were misleading

MickL:
Maybe for swagger the imports are working (i dont know), but for using with validation pipe it was not:

```
export class UpdateUserDto extends PartialType(UserDto) {}
```

Or at least the documentation is kind of misleading because for me it seemed like this page is for how to write a CRUD controller.

Further the types from `@nestjs/mapped-types` should be documented within https://docs.nestjs.com/techniques/validation imo. Actually https://docs.nestjs.com/openapi/mapped-types should be copied nearly 1to1 to https://docs.nestjs.com/techniques/validation?

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[X] Docs
[ ] Other... Please describe:
```

```
[ ] Yes
[X] No
```

### DIFF
--- a/content/openapi/mapped-types.md
+++ b/content/openapi/mapped-types.md
@@ -31,7 +31,7 @@ By default, all of these fields are required. To create a type with the same fie
 export class UpdateCatDto extends PartialType(CreateCatDto) {}
 ```
 
-> info **Hint** The `PartialType()` function is imported from the `@nestjs/swagger` package.
+> info **Hint** The `PartialType()` function is imported from the `@nestjs/mapped-types` package.
 
 #### Pick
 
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
> Further the types from @nestjs/mapped-types should be documented within https://docs.nestjs.com/techniques/validation imo. 

Yeah, this is definitely missing in the current docs. Would you like to create a PR for this?

As for this PR specifically, you must use `@nestjs/swagger` (instead of `@nestjs/mapped-types`) when you're using `@nestjs/swagger` package to generate OpenAPI docs.

MickL:
How would you put this into the docs? As far as I see this would be a 90% copy of the page above. But wouldnt this be redundant? Also the controllers page is very huge already. Can we make both pages into one maybe?

kamilmysliwiec:
> How would you put this into the docs? As far as I see this would be a 90% copy of the page above. 

Yeah I think we should at the very least copy & paste this into the `Validation` chapter (by the end of the current doc) https://docs.nestjs.com/techniques/validation and just map `@nestjs/swagger` imports to `@nestjs/mapped` types + add a huge warning that if one is using `@nestjs/swagger`, then navigate [here](https://docs.nestjs.com/openapi/mapped-types) and when GraphQL, then [here](https://docs.nestjs.com/graphql/mapped-types)

Issue #1790, opened 2021-02-20
all content is not visible because of the banner


Arhell:
![2](https://user-images.githubusercontent.com/26163841/108607647-41d50680-73ca-11eb-8c9d-5349ff408dd8.jpg)

kamilmysliwiec:
Let's track this here https://github.com/nestjs/docs.nestjs.com/pull/1793

PR #1793, opened 2021-02-21
fix carbon

Arhell:
fix issue https://github.com/nestjs/docs.nestjs.com/issues/1790

### DIFF
--- a/src/app/homepage/homepage.component.ts
+++ b/src/app/homepage/homepage.component.ts
@@ -96,6 +96,7 @@ export class HomepageComponent implements OnInit, OnDestroy, AfterViewInit {
       '.newsletter-wrapper',
     );
     const carbonRef = nativeElement.querySelector('#carbonads');
+    const sponsorsWrapper = nativeElement.querySelector('.sponsors-wrapper');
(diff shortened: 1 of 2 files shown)


Arhell:
cc @kamilmysliwiec

src/app/homepage/homepage.component.ts
```diff
@@ -105,6 +106,10 @@ export class HomepageComponent implements OnInit, OnDestroy, AfterViewInit {
       return;
     }
 
+    if (carbonRef) {
+      sponsorsWrapper.classList.add('sponsors-carbon');
```
kamilmysliwiec, line 110: When is this class removed (upon what conditions)? Like, what if you resize a window?

Arhell: If there is a banner then it is there and it appears, if not then no

Arhell: after fix, before in issue image https://github.com/nestjs/docs.nestjs.com/issues/1790

kamilmysliwiec:
LGTM

PR #1851, opened 2021-03-17
Update nats.md

MickL:
nats.js 2.0 has been released and is not compatible with pervious version. To use NATS with Nest.js nats.js 1.x needs to be installed.


kamilmysliwiec:
LGTM

PR #1852, opened 2021-03-18
Update nats.md

MickL:
I added options that are kind of crucial:

- `waitOnFirstConnect: true` will prevent the microservice from crashing if there is initially no connection. For the client this is even more crucial because if there is initially no connection the app is running but it will never attempt to reconnect
- `maxReconnectAttempts: -1` for infinite reconnect, default is `10`

Both options are the same at nats.js 2.x


kamilmysliwiec:
If these are not the default options, why should we enforce/suggest using them even in the most basic example? Perhaps instead of doing that, we should just add a hint/warning (blockquote).

PR #1853, opened 2021-03-18
Update helmet.md

MickL:
Old import was not working anymore, error:

> Argument of type 'typeof import("/.../node_modules/fastify-helmet/index")' is not assignable to parameter of type 'FastifyPluginCallback<any, Server> | FastifyPluginAsync<any, Server> | Promise<{ default: FastifyPluginCallback<any, Server>; }> | Promise<...>'.

Also I added an `await` since `.register()` returns a promise.

fastify-helmet: 5.3.0
helmet: 4.4.1
@nestjs/platform-fastify: 7.6.14 & 8.0.0-alpha.3

### DIFF
--- a/content/security/helmet.md
+++ b/content/security/helmet.md
@@ -17,7 +17,7 @@ Once the installation is complete, apply it as a global middleware.
 ```typescript
 import * as helmet from 'helmet';
 // somewhere in your initialization file
-app.use(helmet());
+await app.use(helmet());
 ```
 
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
According to the `fastify-helmet` package's typings https://github.com/fastify/fastify-helmet/blob/master/index.d.ts, the default export === `fastifyHelmet` const. I believe your issue is caused by the `esModuleInterop` property enabled in your `tsconfig` file, I might be wrong though.

MickL:
I tested within a new nest project created by nest cli. So all ts-config settings were default, no esModuleInterop.

Tony133 (approved):

PR #1881, opened 2021-04-06
docs(providers): changes part about main idea of provider.

AvantaR:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[x] Docs
[ ] Other... Please describe:
```

As provider is decorated with @Injectable() decorator, it's more important that provider can be injected, not that it can inject other dependencies (because, i.e. controllers can do it as well).

## What is the new behavior?

I propose this small change, which can make this part of providers description more clear.

`The main idea of a provider is that it can inject dependencies` to `The main idea of a provider is that it can be injected as  dependency`

```
[ ] Yes
[X] No
```

### DIFF
--- a/content/components.md
+++ b/content/components.md
@@ -1,6 +1,6 @@
 ### Providers
 
-Providers are a fundamental concept in Nest. Many of the basic Nest classes may be treated as a provider – services, repositories, factories, helpers, and so on. The main idea of a provider is that it can **inject** dependencies; this means objects can create various relationships with each other, and the function of "wiring up" instances of objects can largely be delegated to the Nest runtime system.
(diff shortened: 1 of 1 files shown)


jmcdo29:
I think I need to just sit down and write out an explanation of what the `@Injectable()` decorator is for to add to the docs. Cause the really important thing about it, is that it allows things to be injected into it. A `provider` is generally anything added to the `providers` array, services, classes, option objects, etc. And enhancers (pipes, guards, filters, and interceptors) get this `@Injectable()` decorator, but aren't quite `providers` under normal circumstances. So really, we need to update the docs to go into detail about why this decorator is important, and what a provider really is. Maybe under some `advanced` page or something, cause it gets into some Typescript details that not everyone "just gets".

AvantaR:
@jmcdo29 Great, I think that section really need it! Regards

kamilmysliwiec:
LGTM

PR #1909, opened 2021-05-02
docs(validation): add input validation type decorator example

somehowchris:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[x] Docs
[ ] Other... Please describe:
```

No docs about `@Type` on input validation

## What is the new behavior?
Adds a basic example on using `@Type` to transform inputs with `@Query` and `@Param`

```
[ ] Yes
[x] No
```

## Other information

There is currency a issue open due to `class-transformer` not correctly transforming boolean inputs if they are strings.

### DIFF
--- a/content/techniques/validation.md
+++ b/content/techniques/validation.md
@@ -254,6 +254,31 @@ findOne(
 
 > info **Hint** The `ParseIntPipe` and `ParseBoolPipe` are exported from the `@nestjs/common` package.
 
+If you are using a class to capture more than one parameter or query you can use the `@Type` decorator from `class-transformer` to transform the inpu
+
+```typescript
+class QueryDTO {
+  @IsBoolean()
+  @Type(() => Boolean)
+  public sort: boolean;
+}
+
+class ParamDTO {
+  @IsInt()
+  @Type(() => Number)
+  public id: number;
+}
+
+
+@Get(':id')
+findOne(
+  @Param() params: ParamDTO,
+  @Query() query: QueryDTO,
+) {
+  return 'This action returns a user';
+}
+```
+
 #### Mapped types
 
 As you build out features like **CRUD** (Create/Read/Update/Delete) it's often useful to construct variants on a base entity type. Nest provides several utility functions that perform type transformations to make this task more convenient.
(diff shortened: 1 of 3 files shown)


content/techniques/validation.md
```diff
@@ -254,6 +254,31 @@ findOne(
 
 > info **Hint** The `ParseIntPipe` and `ParseBoolPipe` are exported from the `@nestjs/common` package.
 
+Using a class to capture more than one parameter or query you can use `@Type` of `class-transformer` to transform the input.
```
jmcdo29, line 257:
```suggestion
If you are using a class to capture more than one parameter or query you can use the `@Type` decorator from `class-transformer` to transform the input.
```

jmcdo29:
Other than my comment, LGTM. Thanks for the update.

jmcdo29 (approved):

content/techniques/validation.md
```diff
@@ -254,6 +254,31 @@ findOne(
...
+
+
+@Get(':id')
+findOne(
+  @Param() params: ParamDTO,
+  @Query() query: QueryDTO,
+) {
+  return 'This action returns a user';
+}
+```
```
kamilmysliwiec, line 280: To make this work properly, you would have to turn on the `transform` configuration in the `ValidationPipe`, which is disabled by default.

jmcdo29: This will still allow the `IsBoolean` and `IsInt` validations to run inside the pipe though. Regardless of if `transform: true` is set or not, cause we do call `plainToClass` and these `@Transform()` decorators will be handled then

kamilmysliwiec: What I'm trying to say is that in case `transform` is disabled and there are no other configurations passed into the `ValidationPipe` (`class-validator` specific), the pipe will eventually return the original object (without any transformations applied, see line 133 in `validation.pipe.ts`) so saying:

> If you are using a class to capture more than one parameter or query you can use the `@Type` decorator from `class-transformer` to transform the input

Is partially incorrect, since with the default options the controller won't receive the transformer input, but rather the original one.

jmcdo29: Ah, gotcha, definitely correct in that case. Maybe we can add the "if `transform: true` is set for the pipe options" (or something similar )to the sentance

somehowchris:
Sry, I had a major setback due to an ambulant hospitalization. I will update the PR next week as I will have more energy and time.

PR #1926, opened 2021-05-19
Update caching.md

bodolawale:
await getting cached value

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] Docs
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[ ] No
```

### DIFF
--- a/content/techniques/caching.md
+++ b/content/techniques/caching.md
@@ -41,7 +41,7 @@ constructor(@Inject(CACHE_MANAGER) private cacheManager: Cache) {}
 The `get` method on the `Cache` instance (from the `cache-manager` package) is used to retrieve items from the cache. If the item does not exist in the cache, an exception will be thrown.
 
 ```typescript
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
I think this depends on what storage you use(?). By default, `await` shouldn't be required (correct me if I'm wrong).

bodolawale:
> I think this depends on what storage you use(?). By default, `await` shouldn't be required (correct me if I'm wrong).

It does return a promise by default. from Cache declaration. `get<T>(key: string): Promise<T | undefined>;`.
Unless you are using a callback then it returns void. `get<T>(key: string, callback: (error: any, result: T | undefined) => void): void;`

While working with the `cache-manager` package. I didn't use any storage. I just followed the documentation

kamilmysliwiec:
@bodolawale makes sense, thanks for the explanation! LGTM

PR #1934, opened 2021-05-30
Update GraphQL section of authentication.md to return correct body

elisaado:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [X] Docs

With the current code
```js
return ctx.getContext().req;```

the body will contain the plain graphql query, the guard can't parse it and will always return "unauthorised".


## What is the new behavior?

With the new code, 
```js
return { body: ctx.getArgs() };
```

The getRequest function will return a body with the actual arguments passed in on the graphql query. This will result in the guard being able to validate them.

- [X] No


elisaado:
Nevermind, in the docs it was for the JWT strategy where the graphql arguments are not needed, I was using it in the local strategy.

PR #1941, opened 2021-06-24
Improper example codes.

Youmoo:
Since the doc is about *Standalone applications*, it's more accurate to use `createApplicationContext()` method.

## PR Type

[x] Docs

[x] No


kamilmysliwiec:
Great catch, thanks!

PR #2010, opened 2021-08-18
docs(streaming-files): show how to set content-type and content-dispo…

eamonnmcevoy:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

StreamableFIle example does not show how to set content-type or content-disposition

Issue Number: 2009

## What is the new behavior?
StreamableFIle example improved

- [ ] Yes
- [x] No

### DIFF
--- a/content/techniques/streaming-files.md
+++ b/content/techniques/streaming-files.md
@@ -46,7 +46,7 @@ export class FileController {
 }
 ```
 
-The default content type is `application/octet-stream`, if you need to customize the response you can use `res.set`.
+The default content type is `application/octet-stream`, if you need to customize the response you can use the `res.set` method.
 
 > info **hint** If the @Response parameter is present, it must have `{{ '{' }} passthrough: true {{ '}' }}`.
 
--- a/content/techniques/streaming-files.md
+++ b/content/techniques/streaming-files.md
@@ -31,7 +31,7 @@ Fastify, by default, can support sending files without needing to call `stream.p
 
 You can find a simple example of returning the `package.json` as a file instead of a JSON below, but the idea extends out naturally to images, documents, and any other file type.
 
-> info **hint** If the @Response parameter is present, it must have `{ passthrough: true }`.
+> info **hint** If the @Response parameter is present, it must have `{{ '{' }} passthrough: true {{ '}' }}`.
 
 ```ts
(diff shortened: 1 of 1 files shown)


content/techniques/streaming-files.md
```diff
@@ -39,8 +41,13 @@ import { join } from 'path';
 @Controller('file')
 export class FileController {
   @Get()
+  @Response({ passthrough: true }) res,
```
kamilmysliwiec, line 44: I think you meant to add this as a method argument?

kamilmysliwiec:
@jmcdo29 can you take a look 🤞 ?

content/techniques/streaming-files.md
eamonnmcevoy: Yes, updated

jmcdo29 (commented):
Could we add a mention that `res.set` is only necessary if we're trying to set the content type and file name, and otherwise the type will be `application/octet-stream`

Other than that, looks good to me

kamilmysliwiec:
Perhaps we should split this into 2 examples(?). Basically, instead of updating the existing one, add another example below saying that `res.set` + `@Res({ passthrough })` is necessary when [..]

content/techniques/streaming-files.md
```diff
@@ -45,3 +45,26 @@ export class FileController {
   }
 }
 ```
+
+The default content type is `application/octet-stream`, if you need to customize the response you can use `res.set`.
+
+> info **hint** If the @Response parameter is present, it must have `{ passthrough: true }`.
```
kamilmysliwiec, line 51: Build is failing now (we must escape { } chars)

eamonnmcevoy: Done

content/techniques/streaming-files.md
```diff
@@ -45,3 +45,26 @@ export class FileController {
   }
 }
 ```
+
+The default content type is `application/octet-stream`, if you need to customize the response you can use `res.set`.
```
kamilmysliwiec, line 49:
```suggestion
The default content type is `application/octet-stream`, if you need to customize the response you can use the `res.set` method.
```

kamilmysliwiec:
LGTM

PR #2028, opened 2021-08-31
docs(performance): add file name

Zerotask:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

no file name provided

Issue Number: N/A

## What is the new behavior?
file name added

- [ ] Yes
- [x] No

### DIFF
--- a/content/techniques/performance.md
+++ b/content/techniques/performance.md
@@ -43,6 +43,7 @@ bootstrap();
 By default, Fastify listens only on the `localhost 127.0.0.1` interface ([read more](https://www.fastify.io/docs/latest/Getting-Started/#your-first-server)). If you want to accept connections on other hosts, you should specify `'0.0.0.0'` in the `listen()` call:
 
 ```typescript
+@@filename(main)
 async function bootstrap() {
   const app = await NestFactory.create<NestFastifyApplication>(
     AppModule,


Zerotask:
Netlify build is failing with: `ERR! cb() never called!`
This is not related to my changes.

micalevisk:
I believe they didn't added the file name in that snippet on purpose because it's not a full valid code _(as the first one is)_, it just shows a minor change that you'll need to make.

Zerotask:
> it just shows a minor change that you'll need to make.

And in which wile?

It doesn't add much noise to the documentation to clearify that the main.ts file is meant. 
If a snippet would be suitable for different files, then yes, it could be omitted

micalevisk:
> And in which file?

I believe this clear if you have read from top to bottom. Also the snippet is pretty unique (because it has `bootstrap` etc)

---

When I saw file names on code snippets I assume that I can just copy the snippet, create a file with that name and paste it. Maybe using `// on main.ts file` is better

Zerotask:
> When I saw file names on code snippets I assume that I can just copy the snippet

If you read the documentation from top to down, you wouldn't assume this.

In my opinion it's a bad UX to randomly drop the file name from code snippets. But it's not my decision for this project. If you have a different opinion, just close my PR.

micalevisk:
> you wouldn't assume this.

maybe in another case but for this one, we can as they are pretty close to each other. See:

image

> just close my PR.

I'm just a contributor just like you xD

---

> In my opinion it's a bad UX to randomly drop the file name from code snippets

You can think like this:

- when the snippet have a file name, then it means you can copy it to create a working code
- when it doesn't, it is just a snippet to show off something minimal

image

then the UX will be fine, I guess.

Zerotask:
> when it doesn't, it is just a snippet to show off something minimal

That is not true. There are and there was (before I added some) complete files without file names.
That is just your view.

A file name just give you the extra information where this snippet should be placed in. A developer should never copy and paste something without knowing what it does.

But I don't want to start a discussion here. You have pointed out your view, I did mine.

PR #2053, opened 2021-09-21
docs(openapi): handle mapped-types and add build steps

### LINKED ISSUE #1959: Documentation improvement for Swagger Cli Plugin chapter
## Is your feature request related to a problem? Please describe.
    Using swagger with the CLI plugin has been quite inconsistent for me. I have done multiple projects with @nestjs, but the CLI plugin would behave inconsistently or even not generate the types in one project. It would behave differently depending on which point in a project was it installed (the later the worse it was).
    After dedicating hours to testing current and new nestjs apps, i have found the issues that caused the cli plugin to behave this way, and would like these 2 issues described below to be in the documentation of the clip plugin page, as for none would face these issues again.

## Describe the solution you'd like
Clear documentation in the CLI plugin chapter that 
* when using the cli plugin, you must make sure that the partial types (update dtos) import '@nestjs/swagger' and not '@nestjs/mapped-types'
* running a 'nest build' operation when first installing the plugin, or deleting the dist folder and re-running the build process

## Teachability, Documentation, Adoption, Migration Strategy
* "Please make sure to do a "nest build" operation after modifying the 'nest-cli.json' file. note that you might have to delete the 'dist' folder and then run 'nest build'. "
* "Please make sure that Partial types in the application (like update dto files importing partial-type) import it from '@nestjs/swagger' and not '@nestjs/mapped-types' " 

would save others from facing these issues as they weren't really described anywhere easy to find.

*Note: this feature request is only because i did not find anywhere else i could contribute to the documentation. Please correct me if i am wrong :)

shoaib30:
Closes #1959

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [X] Docs

Issue Number: #1959 

## What is the new behaviour?

Added docs to address the issue as mentioned in issue description

- [ ] Yes
- [X] No

### DIFF
--- a/content/openapi/cli-plugin.md
+++ b/content/openapi/cli-plugin.md
@@ -187,26 +187,6 @@ export interface PluginOptions {
 </table>
 
 Make sure to delete the `/dist` folder and rebuild your application whenever plugin options are updated.
-
-```bash
-$ nest build
-```
-
-Or you can configure `package.json` with the following scripts
-
-```json
-"prebuild": "rimraf dist",
-"build": "nest build",
-```
-
-and run it using
-
-```bash
-$ npm run build
-```
-
-> info **Hint** Setting `"deleteOutDir": true` in  [Global Compiler Options](https://docs.nestjs.com/cli/monorepo#global-compiler-options) will remove the compilation output directory everytime `nest build` is invoked
-
 If you don't use the CLI but instead have a custom `webpack` configuration, you can use this plugin in combination with `ts-loader`:
 
 ```javascript
--- a/content/openapi/cli-plugin.md
+++ b/content/openapi/cli-plugin.md
@@ -53,7 +53,7 @@ export class CreateUserDto {
 }
 ```
 
-> info **Hint** When using `PartialType, OmitType, PickType, IntersectionType` in dtos import them from `@nestjs/swagger` instead of `@nestjs/mapped-types` for the plugin to pick up the schema
+> info **Hint** When using [mapped types utilities](https://docs.nestjs.com/openapi/mapped-types) (like `PartialType`) in DTOs import them from `@nestjs/swagger` instead of `@nestjs/mapped-types` for the plugin to pick up the schema.
 
 The plugin adds appropriate decorators on the fly based on the **Abstract Syntax Tree**. Thus you won't have to struggle with `@ApiProperty` decorators scattered throughout the code.
 
--- a/content/openapi/cli-plugin.md
+++ b/content/openapi/cli-plugin.md
@@ -186,7 +186,7 @@ export interface PluginOptions {
   </tr>
 </table>
 
-On the first run and when ever plugin options are changed. Delete `/dist` folder and build the application using
+Make sure to delete the `/dist` folder and rebuild your application whenever plugin options are updated.
 
 ```bash
 $ nest build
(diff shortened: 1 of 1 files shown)


content/openapi/cli-plugin.md
```diff
@@ -184,6 +186,25 @@ export interface PluginOptions {
...
+On the first run and when ever plugin options are changed. Delete `/dist` folder and build the application using
+
+```bash
+$ nest build
+```
+
+Or you can configure `package.json` with the following scripts
+
+```json
+"prebuild": "rimraf dist",
```
micalevisk, line 198: Instead of using `rimraf` direclty we can delegate this to Nest CLI

`nest-cli.json` must have

```json
"compilerOptions": {
  "deleteOutDir": true
}
```

shoaib30: A nest app created with Neset CLI using `nest new` generates a package.json with `rimraf` in `prebuild` using that made sense to me. 

The `dist` directory needs to be deleted only when changes to `nest-cli.json` are made and not every time a build is done as is the case with `"deleteOutDir": true` 

Does removing the `dist` directory every time have any impact on the build performance? If not then your suggestion is a lot better to do.

micalevisk: > Does removing the `dist` directory every time have any impact on the build performance?

Since the output dir will be removed by Nest CLI only upon `nest build` then both solutions will perform the same way I guess.

I'd say: keep your solution. But it would be cool to also add something like "Or using `"deleteOutDir": true` on [global compiler options](https://docs.nestjs.com/cli/monorepo#global-compiler-options)"

shoaib30: sounds good, I've added it

micalevisk (approved):

content/openapi/cli-plugin.md
```diff
@@ -53,6 +53,8 @@ export class CreateUserDto {
 }
 ```
 
+> info **Hint** When using `PartialType, OmitType, PickType, IntersectionType` in dtos import them from `@nestjs/swagger` instead of `@nestjs/mapped-types` for the plugin to pick up the schema
```
micalevisk, line 56:
```suggestion
> info **Hint** When using [mapped types utilities](https://docs.nestjs.com/openapi/mapped-types) (like `PartialType`) in DTOs import them from `@nestjs/swagger` instead of `@nestjs/mapped-types` for the plugin to pick up the schema.
```

content/openapi/cli-plugin.md
```diff
@@ -184,6 +186,27 @@ export interface PluginOptions {
   </tr>
 </table>
 
+On the first run and when ever plugin options are changed. Delete `/dist` folder and build the application using
```
kamilmysliwiec, line 189:
```suggestion
Make sure to delete  the `/dist` folder and rebuild your application  whenever plugin options are updated.
```

content/openapi/cli-plugin.md
```diff
@@ -184,6 +186,27 @@ export interface PluginOptions {
...
+```
+
+and run it using
+
+```bash
+$ npm run build
+```
+
+> info **Hint** Setting `"deleteOutDir": true` in  [Global Compiler Options](https://docs.nestjs.com/cli/monorepo#global-compiler-options) will remove the compilation output directory everytime `nest build` is invoked
+
```
kamilmysliwiec, line 209:
```suggestion

```

PR #2101, opened 2021-11-10
docs: @Headers() to @Header()

chanjsq:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/content/controllers.md
+++ b/content/controllers.md
@@ -132,7 +132,7 @@ The request object represents the HTTP request and has properties for the reques
       <td><code>req.query</code> / <code>req.query[key]</code></td>
     </tr>
     <tr>
-      <td><code>@Headers(name?: string)</code></td>
+      <td><code>@Header(name?: string)</code></td>
       <td><code>req.headers</code> / <code>req.headers[name]</code></td>
     </tr>
     <tr>


content/controllers.md
```diff
@@ -132,7 +132,7 @@ The request object represents the HTTP request and has properties for the reques
       <td><code>req.query</code> / <code>req.query[key]</code></td>
     </tr>
     <tr>
-      <td><code>@Headers(name?: string)</code></td>
+      <td><code>@Header(name?: string)</code></td>
```
jmcdo29, line 135: `@Header()` is a method decorator, for telling Nest to add a specific header to the response. `@Headers()` is a parameter decorator, to tell  Nest to retrieve `req.header` or `req.headers[name]`

jmcdo29 (commented):
The current docs are correct.

content/controllers.md
chanjsq: Thank you for your detailed and quick explanation. I confused them. But maybe they are easy to confuse. Since `@Headers()` is not consistent with `@Query()`, `@Param()`, the latters are singular, but the former is plural. Am I right?

jmcdo29: Well, `req.query` is what `@Query()` relates to. `req.params` is `@Param()`, but there's no setting of url parameters like there is of response headers. Express and Fastify's API for the `response` object use `header` for setting a header and for the request object use `headers` for retrieving the headers of the incoming request, so we're honestly just following the example they have.

chanjsq: I consult the Fastify API, the `request` object did use `headers` to retrieve headers from the incoming request, but it also use `params` to retrieve the path parameters. They keep the same case, so they are consistent in Fastify.

refs:
- [Fastify Request API](https://www.fastify.io/docs/latest/Request/)
- [Fastify Reply API](https://www.fastify.io/docs/latest/Reply/)

jmcdo29: That's what I said, isn't it? `header` for setting, `headers` for getting.

`@Param()` maps to `req.params`, which is different, yes, but there's also no way to set URL parameters on a response, so it's a moot point. `@Param()` and `@Params()` would technically be referencing the same thing.

chanjsq: Oh, I got it. Thank you very much for your patience! 👍 I just started learning NestJS yesterday, and bought your official course, it was very helpful. NestJS is very awesome, I will prompt it to my friends in China. Thank you for your awesome work!

PR #2107, opened 2021-11-13
doc: replace 'catch everything' example with a platform-agnostic one

micalevisk:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

There aren't examples in the docs regarding using the `HttpAdapter` interface to write a platform-agnostic code

<details>
<summary>current version</summary>

image

</details>

## What is the new behavior?

I think the ['catch everything' interceptor example](https://docs.nestjs.com/exception-filters#catch-everything) is a good case on how we can write a platform-agnostic code. And also because the old code is almost equal to the [first example](https://docs.nestjs.com/exception-filters#catch-everything) so there wasn't much learning of the capabilities of NestJS (like being platform-agnostic) after reading both examples in there. Thus I changed it to:

image

- [ ] Yes
- [x] No

### DIFF
--- a/content/exception-filters.md
+++ b/content/exception-filters.md
@@ -284,6 +284,8 @@ You can add as many filters with this technique as needed; simply add each to th
 
 In order to catch **every** unhandled exception (regardless of the exception type), leave the `@Catch()` decorator's parameter list empty, e.g., `@Catch()`.
 
+In the example below we have a code that is platform-agnostic because it uses the [HTTP adapter](./faq/http-adapter) to deliver the response, and doesn't use any of the platform-specific objects (`Request` and `Response`) directly:
+
 ```typescript
 import {
   ExceptionFilter,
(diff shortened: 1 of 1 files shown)


content/exception-filters.md
```diff
@@ -292,30 +294,35 @@ import {
...
 } from '@nestjs/common';
+import { AbstractHttpAdapter, HttpAdapterHost } from '@nestjs/core';
 
 @Catch()
 export class AllExceptionsFilter implements ExceptionFilter {
-  catch(exception: unknown, host: ArgumentsHost) {
+  private readonly httpAdapter: AbstractHttpAdapter;
+
+  constructor(httpAdapterHost: HttpAdapterHost) {
+    this.httpAdapter = httpAdapterHost.httpAdapter;
```
kamilmysliwiec, line 304: In certain situations, `httpAdapter` might not be yet available in a constructor so I'd suggest retrieving it within the `catch` method instead.

micalevisk: good to know.

image

kamilmysliwiec:
LGTM

PR #2141, opened 2021-12-19
docs(filters): fix passing error parameter

olivewind:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

Issue Number: N/A

- [ ] Yes
- [x] No

## Other information

image

### DIFF
--- a/content/exception-filters.md
+++ b/content/exception-filters.md
@@ -364,8 +364,8 @@ The first method is to inject the `HttpServer` reference when instantiating the 
 async function bootstrap() {
   const app = await NestFactory.create(AppModule);
 
-  const { httpAdapter } = app.get(HttpAdapterHost);
-  app.useGlobalFilters(new AllExceptionsFilter(httpAdapter));
+  const httpAdapterHost = app.get(HttpAdapterHost);
+  app.useGlobalFilters(new AllExceptionsFilter(httpAdapterHost));
 
   await app.listen(3000);
 }


kamilmysliwiec:
HttpAdapterHost != ArgumentsHost

olivewind:
@kamilmysliwiec I know, but according to the code provided by the documentation, an error will be thrown

jmcdo29:
I think the screenshot is hghlighting the wrong thing here. The code change, I think, is correct though. The `AllExceptionsFilter` in the docs does expect an instance of `HttpAdapterHost` passed to the constructor, but in the code snippet we are pulling out the `httpAdapter` from the `HttpAdapterHost`

```ts
@Catch()
  constructor(private readonly httpAdapterHost: HttpAdapterHost) {}

  catch(exception: unknown, host: ArgumentsHost): void {
    // In certain situations `httpAdapter` might not be available in the
    // constructor method, thus we should resolve it here.
    const { httpAdapter } = this.httpAdapterHost;

    const ctx = host.switchToHttp();

    const httpStatus =
      exception instanceof HttpException
        ? exception.getStatus()
        : HttpStatus.INTERNAL_SERVER_ERROR;

    const responseBody = {
      statusCode: httpStatus,
      timestamp: new Date().toISOString(),
      path: httpAdapter.getRequestUrl(ctx.getRequest()),
    };

    httpAdapter.reply(ctx.getResponse(), responseBody, httpStatus);
  }
}
```

```ts

async function bootstrap() {

  const { httpAdapter } = app.get(HttpAdapterHost);
  app.useGlobalFilters(new AllExceptionsFilter(httpAdapter));

  await app.listen(3000);
}
bootstrap();
```

kamilmysliwiec:
@jmcdo29 I think it depends on what example we're looking at. The latter one with `AllExceptionsFilter extends BaseExceptionFilter` requires the `HttpServer` while the "custom implementation" needs the `HttpAdapterHost`. We might need to update these examples to make it more consistent

PR #2145, opened 2021-12-24
Fix npm installation command

saeedvaziry:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/content/techniques/caching.md
+++ b/content/techniques/caching.md
@@ -7,8 +7,8 @@ Caching is a great and simple **technique** that helps improve your app's perfor
 First install required packages:
 
 ```bash
-$ npm install cache-manager
-$ npm install -D @types/cache-manager
+$ npm install --save cache-manager
+$ npm install --save-dev @types/cache-manager
 ```
 
 #### In-memory cache


content/techniques/caching.md
```diff
@@ -7,8 +7,8 @@ Caching is a great and simple **technique** that helps improve your app's perfor
 First install required packages:
 
 ```bash
-$ npm install cache-manager
-$ npm install -D @types/cache-manager
+$ npm install --save cache-manager
+$ npm install --save-dev @types/cache-manager
```
jmcdo29, line 11: `--save` is the default behavior as on npm 5, so there's no need to add it. `-D` is shorthand for `--save-dev` so I don't see the need to change this.

saeedvaziry: @jmcdo29 Yes you are right. The only purpose of this PR is to make it similar to other places in the documents

kamilmysliwiec: @saeedvaziry we're not 💯 % consistent but since it's not a high priority, we'll likely just update commands one at a time.

PR #2147, opened 2021-12-26
docs(): page link improvement part one 

Tony133:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

Issue Number: N/A

## What is the new behavior?
Remove warning `Dangling Links Found`

- [ ] Yes
- [x] No

## Other information
This PR, Remove warning `Dangling Links Found`(see screenshot)

I split this improvement into 3 PRs to do a little faster review.

screenshot

### DIFF
--- a/content/cli/usages.md
+++ b/content/cli/usages.md
@@ -70,7 +70,7 @@ $ nest g <schematic> <name> [options]
 | `pipe`        | `pi`  | Generate a pipe declaration.                                                                        |
 | `provider`    | `pr`  | Generate a provider declaration.                                                                    |
 | `resolver`    | `r`   | Generate a resolver declaration.
-| `resource`    | `res` | Generate a new CRUD resource. See the [CRUD (resource) generator](/recipes/crud-generator) for more details.                                                                    |
+| `resource`    | `res` | Generate a new CRUD resource. See the [CRUD (resource) generator](https://docs.nestjs.com/recipes/crud-generator) for more details.                                                                    |
 | `service`     | `s`   | Generate a service declaration.                                                                    |
 
 ##### Options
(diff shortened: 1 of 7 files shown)


content/cli/usages.md
```diff
@@ -70,7 +70,7 @@ $ nest g <schematic> <name> [options]
 | `pipe`        | `pi`  | Generate a pipe declaration.                                                                        |
 | `provider`    | `pr`  | Generate a provider declaration.                                                                    |
 | `resolver`    | `r`   | Generate a resolver declaration.
-| `resource`    | `res` | Generate a new CRUD resource. See the [CRUD (resource) generator](/recipes/crud-generator) for more details.                                                                    |
+| `resource`    | `res` | Generate a new CRUD resource. See the [CRUD (resource) generator](https://docs.nestjs.com/recipes/crud-generator) for more details.                                                                    |
```
kamilmysliwiec, line 73: What's the reason for this change?

kamilmysliwiec: (this question applies to most of the updates in this PR)

Tony133: I did these PRs, to remove the warnings that occur when we do `npm run build` (see screenshot)

To remove them I added this part `https://docs.nestjs.com`to each url where there was the warning.
It should always be added to have no warnings in the future.

screenshot

kamilmysliwiec: The local version of the application won't be working as expected now as every link will route you to the deployed version of the app (instead of the local one running on your machine) making the development process quite frustrating

Tony133: Got it, can this be fixed when a new version of NestJS (v9) comes out by updating the site?

kamilmysliwiec: Not sure if I'm following(?). Explicitly adding `docs.nestjs.com` to the link just makes no sense as the application won't run as expected between different environments.

Tony133: Understood , In many URLs on the site this is already there (see screenshot) but in others it is missing and generates the warning.

screensho2

Tony133: These are the same warnings that you also find in the Netfily console (see screenshot)

screenshot3

kamilmysliwiec: > Understood, In many URLs on the site this is already there (see screenshot) but in others it is missing and generates the warning.

We use `docs.nestjs.com` explicitly **only** if we want a link to be opened in a new tab

Tony133: Ok, so we leave the warnings like this? 
You can also find them in the Netfily console (you saw the screenshot above)? 

However it would be bad for each link to open a new tab, it is true that it depends on the cases, but for a guide it would be interesting, so you can pick up where you were reading.

kamilmysliwiec: > Ok, so we leave the warnings like this?

yeah these warnings are acceptable ✅

Tony133: Ok 👍, if you want these links to open in another tab ok, otherwise close these PRs as well. Then if you want to think about it let me know, no problem for me. 😉 👍

PR #2176, opened 2022-01-20
docs(security): fix req type of login example

kaeltab:
This fixes the following error: Parameter 'req' implicitly has an 'any' type.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

nestjs-requestdecorator-error2

Issue Number: N/A

## What is the new behavior?

No typing error.

- [ ] Yes
- [x] No

### DIFF
--- a/content/security/authentication.md
+++ b/content/security/authentication.md
@@ -313,14 +313,15 @@ Open the `app.controller.ts` file and replace its contents with the following:
 
 ```typescript
 @@filename(app.controller)
-import { Controller, Request, Post, UseGuards } from '@nestjs/common';
+import { Controller, Req, Post, UseGuards } from '@nestjs/common';
 import { AuthGuard } from '@nestjs/passport';
+import { Request } from 'express';
 
 @Controller()
 export class AppController {
   @UseGuards(AuthGuard('local'))
   @Post('auth/login')
-  async login(@Request() req) {
+  async login(@Req() req: Request) {
     return req.user;
   }
 }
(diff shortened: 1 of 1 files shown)


micalevisk (approved):

content/security/authentication.md
```diff
@@ -313,14 +313,15 @@ Open the `app.controller.ts` file and replace its contents with the following:
...
+import { Controller, Req, Post, UseGuards } from '@nestjs/common';
 import { AuthGuard } from '@nestjs/passport';
+import { Request } from 'express';
 
 @Controller()
 export class AppController {
   @UseGuards(AuthGuard('local'))
   @Post('auth/login')
-  async login(@Request() req) {
+  async login(@Req() req: Request) {
```
jmcdo29, line 324: The `Request` here is either `Request` from the `express` package or `fastifyRequest` from the `fastify` package, depending on your underlying HTTP engine. `Request` from `@nestjs/common` is an alias for the `@Req()` decorator so you can do `@Request()` or `@Req()`

jmcdo29 (changes requested):

content/security/authentication.md
micalevisk: the "issue" OP's trying to fix here is the `any` type of `req`, which can lead to TS errors when we create a new nestjs app with `strict` option (making `noImplicitAny: true`)

micalevisk: image

maybe adding a comment in there to the fastify version would be good because this page has no reference to fastify

bikeshedder: > I get that, but `Request` from `@nestjs/common` is not the right type here.

The PR never used `Request` from `@nestjs/common`:
```ts
import { Request } from 'express';
```

I don't get why this PR wasn't merged. Any fastify user should know what to do here. I had a hard time finding out what type that request object is supposed to be.

PR #2183, opened 2022-01-22
add ApiPropertyOptional

tahamansoor:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [x] Docs

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/content/openapi/types-and-parameters.md
+++ b/content/openapi/types-and-parameters.md
@@ -34,6 +34,16 @@ export class CreateCatDto {
 
 > info **Hint** Instead of manually annotating each property, consider using the Swagger plugin (see [Plugin](/openapi/cli-plugin) section) which will automatically provide this for you.
 
+use `@ApiPropertyOptional()` for Optional Parameters
+
+```typescript
+import { ApiPropertyOptional} from '@nestjs/swagger';
+
+export class CreateCatDto {
+  @ApiPropertyOptional()
+  breed: string;
+}
+```
 Let's open the browser and verify the generated `CreateCatDto` model:
 
 <figure><img src="/assets/swagger-dto2.png" /></figure>


content/openapi/types-and-parameters.md
```diff
@@ -34,6 +34,16 @@ export class CreateCatDto {
...
 > info **Hint** Instead of manually annotating each property, consider using the Swagger plugin (see [Plugin](/openapi/cli-plugin) section) which will automatically provide this for you.
 
+use `@ApiPropertyOptional()` for Optional Parameters
+
+```typescript
+import { ApiPropertyOptional} from '@nestjs/swagger';
+
+export class CreateCatDto {
+  @ApiPropertyOptional()
+  breed: string;
```
micalevisk, line 44:
```suggestion
  breed?: string;
```

micalevisk:
What if you add a new optional field on the `CreateCatDto` that already exists? It will be easier to follow I guess

tahamansoor:
> What if you add a new optional field on the `CreateCatDto` that already exists? It will be easier to follow I guess

` name?: string;` 
are you saying like mentioned above? if yes that does not work I have tried this before

micalevisk:
nvm

I think if you just change the line

image

to

```
As you can see, the definition is empty although the class has a few declared properties.
In order to make the class properties visible to the `SwaggerModule`, we have to either annotate them with the `@ApiProperty()` decorator (or `@ApiOptionalProperty()` for optional fields) or use the CLI plugin (read more in the Plugin section) which will do it automatically:
```

would be enough

tahamansoor:
> nvm
> 
> I think if you just change the line
> 
> image
> 
> to
> 
> ```
> As you can see, the definition is empty although the class has a few declared properties.
> In order to make the class properties visible to the `SwaggerModule`, we have to either annotate them with the `@ApiProperty()` decorator (or `@ApiOptionalProperty()` for optional fields) or use the CLI plugin (read more in the Plugin section) which will do it automatically:
> ```
> 
> would be enough

yes that's a good point

content/openapi/types-and-parameters.md
johannesschobel, line 44: i guess we should use the `?` here?!

```suggestion
  breed?: string;
```

johannesschobel (changes requested):

kamilmysliwiec:
We decided not to mention `@ApiPropertyOptional` and encourage developers to use `@ApiProperty({ required: false })` instead

content/openapi/types-and-parameters.md
tahamansoor: I have tried it 
It doesn't work

PR #2221, opened 2022-02-22
feat: mention the different registration methods

jmcdo29:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [X] Docs

Currently we don't have guidelines or documentation around `forRoot`, `forFeature` and `register`.

## What is the new behavior?

I've added in a segment about forRoot, forFeature, and register. I've seen
this question pop up several times on Discord, in GitHub Issues, and in
StackOverflow. I figured it was time for us to try and codify a reason
for the naming. Of course, there's no hard rule here, but the guideline
is suggested and it could become community standard which would be a big
win for the mental load of creating and using new modules.

- [ ] Yes
- [X] No

### DIFF
--- a/content/fundamentals/dynamic-modules.md
+++ b/content/fundamentals/dynamic-modules.md
@@ -265,11 +265,13 @@ export const CONFIG_OPTIONS = 'CONFIG_OPTIONS';
 
 You may have seen the use for methods like `forRoot`, `register`, and `forFeature` around some of the `@nestjs/` packages and may be wondering what the difference for all of these methods are. There is no hard rule about this, but the `@nestjs/` packages try to follow these guidelines:
 
-When creating a module with `register`, you are expecting to configure a dynamic module with a specific configuration for use only by the calling module . For example, with Nest's `@nestjs/axios`: `HttpModule.register({{ '{' }} baseUrl: 'someUrl' {{ '}' }})`. If, in another module you use `HttpModule.register({{ '{' }} baseUrl: 'somewhere else' {{ '}' }})`, it will have the different configuration. You can do this for as many modules as you want.
+When creating a module with:
 
-When creating a module with `forRoot`, you are expecting to configure a dynamic module once and reuse that configuration in multiple places (though possibly unknowingly as it's abstracted away). This is why you have one `GraphQLModule.forRoot()`, one `TypeOrmModule.forRoot()`, etc. 
+* `register`, you are expecting to configure a dynamic module with a specific configuration for use only by the calling module . For example, with Nest's `@nestjs/axios`: `HttpModule.register({{ '{' }} baseUrl: 'someUrl' {{ '}' }})`. If, in another module you use `HttpModule.register({{ '{' }} baseUrl: 'somewhere else' {{ '}' }})`, it will have the different configuration. You can do this for as many modules as you want.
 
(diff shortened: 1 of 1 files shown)


Tony133 (approved):

ruslanguns (commented):
Amazing!

content/fundamentals/dynamic-modules.md
```diff
@@ -261,6 +261,18 @@ One final note: for simplicity we used a string-based injection token (`'CONFIG_
 export const CONFIG_OPTIONS = 'CONFIG_OPTIONS';
 ```
 
+### Community Guidelines
+
+You may have seen the use for methods like `forRoot`, `register`, and `forFeature` around some of the `@nestjs/` packages and may be wondering what the difference for all of these methods are. There is no hard rule about this, but the `@nestjs/` packages try to follow these guidelines:
+
+When creating a module with `register` you are expecting to configuring a dynamic module for this module only with a specific configuration. For example, with Nest's `@nestjs/axios`: `HttpModule.register({{ '{' }} baseUrl: 'someUrl' {{ '}' }})`. If, in another module you use `HttpModule.register({{ '{' }} baseUrl: 'somewhere else' {{ '}' }})`, it will have the different configuration. You can do this for as many modules as you want.
```
thomas-w-grant, line 268:
```suggestion
When creating a module with `register`, you are expecting to configure a dynamic module with a specific configuration for use only by the calling module . For example, with Nest's `@nestjs/axios`: `HttpModule.register({{ '{' }} baseUrl: 'someUrl' {{ '}' }})`. If, in another module you use `HttpModule.register({{ '{' }} baseUrl: 'somewhere else' {{ '}' }})`, it will have the different configuration. You can do this for as many modules as you want.
```

content/fundamentals/dynamic-modules.md
```diff
@@ -261,6 +261,18 @@ One final note: for simplicity we used a string-based injection token (`'CONFIG_
...
+
+When creating a module with `forRoot`, you are expecting to configure a dynamic module once and re-using that configuration in multiple places (though possibly unknowingly as it's abstracted away). This is why you have one `GraphQLModule.forRoot()`, one `TypeOrmModule.forRoot()`, etc.
```
thomas-w-grant, line 270:
```suggestion
When creating a module with `forRoot`, you are expecting to configure a dynamic module once and reuse that configuration in multiple places (though possibly unknowingly as it's abstracted away). This is why you have one `GraphQLModule.forRoot()`, one `TypeOrmModule.forRoot()`, etc. 
```

thomas-w-grant (changes requested):
From the perspective of a newbie to Nest, let me attest to how valuable this additional documentation is.  Thank you ❤️ 

I have made some suggested changes.  My motivation is two fold.  One, I want to make sure that I am understanding you and my suggestions will hopefully demonstrate that I either understand or don't understand.  Two, I think my feedback provides some minor changes to grammar/spelling.

kamilmysliwiec:
This is a great initiative @jmcdo29, thank you! I'll give it a more thorough review shortly :)

Tony133:
I thought, if we do something like this:

screenshot

not to repeat three times "`When creating a form with ...`"

what do you think?

micalevisk (approved):
image

kamilmysliwiec:
LGTM

PR #2225, opened 2022-02-25
docs: use different decorator in example for PartialType with two parameters

kimroen:
Updates the example for PartialType with a different decorator for the parent to have a different decorator like the text suggests.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/content/graphql/mapped-types.md
+++ b/content/graphql/mapped-types.md
@@ -38,7 +38,7 @@ export class UpdateUserInput extends PartialType(CreateUserInput) {}
 The `PartialType()` function takes an optional second argument that is a reference to a decorator factory.  This argument can be used to change the decorator function applied to the resulting (child) class.  If not specified, the child class effectively uses the same decorator as the **parent** class (the class referenced in the first argument). In the example above, we are extending `CreateUserInput` which is annotated with the `@InputType()` decorator.  Since we want `UpdateUserInput` to also be treated as if it were decorated with `@InputType()`, we didn't need to pass `InputType` as the second argument. If the parent and child types are different, (e.g., the parent is decorated with `@ObjectType`), we would  pass `InputType` as the second argument. For example:
 
 ```typescript
-@InputType()
+@ObjectType()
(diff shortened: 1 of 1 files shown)


content/graphql/mapped-types.md
```diff
@@ -38,7 +38,7 @@ export class UpdateUserInput extends PartialType(CreateUserInput) {}
 The `PartialType()` function takes an optional second argument that is a reference to a decorator factory.  This argument can be used to change the decorator function applied to the resulting (child) class.  If not specified, the child class effectively uses the same decorator as the **parent** class (the class referenced in the first argument). In the example above, we are extending `CreateUserInput` which is annotated with the `@InputType()` decorator.  Since we want `UpdateUserInput` to also be treated as if it were decorated with `@InputType()`, we didn't need to pass `InputType` as the second argument. If the parent and child types are different, (e.g., the parent is decorated with `@ObjectType`), we would  pass `InputType` as the second argument. For example:
```
johannesschobel, line 38:
```suggestion
The `PartialType()` function takes an optional second argument that is a reference to a decorator factory. This argument can be used to change the decorator function applied to the resulting (child) class.  If not specified, the child class effectively uses the same decorator as the **parent** class (the class referenced in the first argument). In the example above, we are extending `CreateUserInput` which is annotated with the `@InputType()` decorator.  Since we want `UpdateUserInput` to also be treated as if it were decorated with `@InputType()`, we didn't need to pass `InputType` as the second argument. If the parent and child types are different, (e.g., the parent is decorated with `@ObjectType`), we would  pass `InputType` as the second argument. For example:
```

johannesschobel (changes requested):

kamilmysliwiec:
Perhaps we should tweak the wording here instead of updating the example here (as the example seems to be valid & correct)

PR #2248, opened 2022-03-17
docs(terminus): replaced AppModule with the HealthModule

Scrip7:

## PR Type

- [x] Docs

The code block for writing the health controller exists, but there isn't a sample regarding its module.

## What is the new behavior?

I've added the code block.

- [x] No

### DIFF
--- a/content/recipes/terminus.md
+++ b/content/recipes/terminus.md
@@ -31,13 +31,9 @@ A health check represents a summary of **health indicators**. A health indicator
 - `MemoryHealthIndicator`
 - `DiskHealthIndicator`
 
-To get started with our first health check, let's generate the `HealthModule` now using [Nest CLI](/cli/overview).
+To get started with our first health check, let's create the `HealthModule` and import the `TerminusModule` into it in its imports array.
 
-```bash
-$ nest g module health
-```
-
-Once generated, we can import the `TerminusModule` into it in its imports array.
+> info **Hint** To create the module using the [Nest CLI](cli/overview), simply execute the `$ nest g module health` command.
 
 ```typescript
 @@filename(health.module)
--- a/content/recipes/terminus.md
+++ b/content/recipes/terminus.md
@@ -22,26 +22,32 @@ $ npm install --save @nestjs/terminus
 
 A health check represents a summary of **health indicators**. A health indicator executes a check of a service, whether it is in a healthy or unhealthy state. A health check is positive if all the assigned health indicators are up and running. Because a lot of applications will need similar health indicators, [`@nestjs/terminus`](https://github.com/nestjs/terminus) provides a set of predefined indicators, such as:
 
-- `HttpHealthIndicator`
-- `TypeOrmHealthIndicator`
-- `MongooseHealthIndicator`
-- `SequelizeHealthIndicator`
-- `MicroserviceHealthIndicator`
-- `GRPCHealthIndicator`
-- `MemoryHealthIndicator`
-- `DiskHealthIndicator`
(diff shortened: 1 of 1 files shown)


micalevisk (changes requested):
instead of adding another module, we could just add that `HealthController` to `AppModule`'s `controllers` array as it is used in the beginning of the page, just to make things more easy to follow.

I guess people can get confused because in this page we already have the `imports: [TerminusModule]`

Scrip7:
Oh, sorry, my bad. @micalevisk 
I knew we could import that in the root `AppModule`, but I didn't notice we did that on top.

Wouldn't it be better if we guided people to create a separate `HealthModule` for that?

Cause I believe in a real-world scenario (at least in the handful scenarios I've had in the past);
- The `HealthController` is also dependent on the `HttpModule` from the `@nestjs/axios` and the configuration services to grab other microservices information
- Other modules shouldn't have access to the set of health indicators `@nestjs/terminus` provides
- It'll help organize health-checker-related things in its module rather than putting everything in the root module.

Please correct me if I'm wrong.

micalevisk:
having `HealthModule` is far better and realistc for sure, but I think we're trying to make this example as simple as possible

actually, I don't know if we need to show the final version of the module because if we follow that page properly, we have ran `nest g controller health` and this will update the `AppModule` for us

:thinking:

Scrip7:
Sure, I'll give it a try!
I was looking at the [samples](https://github.com/nestjs/terminus/tree/master/sample),
All of them have their dedicated health module 😄

content/recipes/terminus.md
```diff
@@ -22,26 +22,32 @@ $ npm install --save @nestjs/terminus
...
-- `MemoryHealthIndicator`
-- `DiskHealthIndicator`
+-   `HttpHealthIndicator`
+-   `TypeOrmHealthIndicator`
+-   `MongooseHealthIndicator`
+-   `SequelizeHealthIndicator`
+-   `MicroserviceHealthIndicator`
+-   `GRPCHealthIndicator`
+-   `MemoryHealthIndicator`
+-   `DiskHealthIndicator`
```
micalevisk, line 32:
```suggestion
- `HttpHealthIndicator`
- `TypeOrmHealthIndicator`
- `MongooseHealthIndicator`
- `SequelizeHealthIndicator`
- `MicroserviceHealthIndicator`
- `GRPCHealthIndicator`
- `MemoryHealthIndicator`
- `DiskHealthIndicator`
```

micalevisk (approved):

content/recipes/terminus.md
```diff
@@ -31,17 +31,23 @@ A health check represents a summary of **health indicators**. A health indicator
 - `MemoryHealthIndicator`
 - `DiskHealthIndicator`
 
-To get started with our first health check, we need to import the `TerminusModule` into our `AppModule`.
+To get started with our first health check, let's generate the `HealthModule` now using [Nest CLI](/cli/overview).
```
kamilmysliwiec, line 34: We can't assume everyone uses CLI. We should skip this sentence and consider adding it as a hint

Scrip7: Are you sure about that?
As Micael said earlier, the `nest g controller health` was there before any changes.

Should I still skip it?

kamilmysliwiec:
@jmcdo29 can you double-check this when you have a chance? 🤞

jmcdo29 (approved):
LGTM

Tony133 (approved):

PR #2252, opened 2022-03-20
docs(first-steps): update nodejs minimum version to v14

micalevisk:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

See https://github.com/nestjs/nest/issues/9354

- [ ] Yes
- [x] No

### DIFF
--- a/content/first-steps.md
+++ b/content/first-steps.md
@@ -10,7 +10,7 @@ We'll mostly use TypeScript in the examples we provide, but you can always **swi
 
 #### Prerequisites
 
-Please make sure that [Node.js](https://nodejs.org/) (>= 10.13.0, except for v13) is installed on your operating system.
+Please make sure that [Node.js](https://nodejs.org/) (>= 14.0.0 or LTS) is installed on your operating system.
 
 #### Setup
 


Tony133 (approved):

kamilmysliwiec:
Nest should still work fine on v10.13.0 (although it's not an LTS version). What's the reason for this (breaking) change?

Tony133:
@kamilmysliwiec, it turned out that with version 10.13 the app crashed [see problem here](https://github.com/nestjs/nest/issues/9354), the problem was solved by updating the version of nodejs and we thought that since version 10.13 which later became LTS see [here] (https://nodejs.org/ru/blog/release/v10.13.0/) its maintenance cycle is finished (end of life) and since nodejs version 12 also ends maintenance next month, we thought about updating the documentation with a version 14 of Nodejs as it is still under maintenance for some time or an LTS version.

This is a brief summary of the why of this PR. 😁

kamilmysliwiec:
I just double-checked and Nest works as expected with Node v10.13.

image

Requirement !== recommendation. 

We recommend everyone to use most recent & stable versions (at least LTS) but this doesn't mean Nest isn't compatible with older releases.

Tony133:
Right, I agree  `Requirement !== recommendation` 👍

I usually always use LTS as the nodeJS version. 😄

I'll let you notice one thing, look at this minimal reproduction `https://github.com/AmrAbdalrahman/test-nest` gives error with 10.13 while going to 16.13.1 it works and look at these screenshots i did a clean installation from scratch with version 10.13.0 and then I switched to 16.13.1 and it only works with 16.13.1.


screenshot

From what I understand the error is only with a clean installation of Nest using nodejs 10.13, but I will do more tests. By any chance it gives you an error while installing clean with nodejs 10.13?

PR #2263, opened 2022-04-02
docs(rate-limiting): improvements in the customization section

Tony133:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

Issue Number: N/A

## What is the new behavior?
Improvements in the customization section for rate-limiting 

- [ ] Yes
- [x] No

## Other information
This PR, introduces improvements in the customization section for speed limiting, as the customization part lacked a "practical" example to better understand the operation, otherwise you must always see the readme of the `@nestjs/throttler` package.

### DIFF
--- a/content/security/rate-limiting.md
+++ b/content/security/rate-limiting.md
@@ -47,7 +47,6 @@ This `@SkipThrottle()` decorator can be used to skip a route or a class or to ne
 @SkipThrottle()
 @Controller('users')
 export class UsersController {
-
   // This route will skip rate limiting.
   @SkipThrottle(false)
   dontSkip() {
--- a/content/security/rate-limiting.md
+++ b/content/security/rate-limiting.md
@@ -47,15 +47,14 @@ This `@SkipThrottle()` decorator can be used to skip a route or a class or to ne
 @SkipThrottle()
 @Controller('users')
 export class UsersController {
-  // This route will skip rate limiting.
+  // Rate limiting is applied to this route.
   @SkipThrottle(false)
   dontSkip() {
-    return "List users work with Rate limiting"
+    return "List users work with Rate limiting.";
   }
-
-  // Rate limiting is applied to this route.
+  // This route will skip rate limiting.
   doSkip() {
-    return "List users work without Rate limiting"
+    return "List users work without Rate limiting.";
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


content/security/rate-limiting.md
```diff
@@ -35,13 +35,48 @@ Once the module has been imported, you can then choose how you would like to bin
...
+export class UsersController {}
+```
+
+This `@SkipThrottle()` decorator can be used to skip a route or a class or to negate the skipping of a route in a class that is skipped.
+
+```typescript
+@SkipThrottle()
+@Controller('users')
+export class UsersController {
+
```
micalevisk, line 50:
```suggestion
```

content/security/rate-limiting.md
```diff
@@ -35,13 +35,48 @@ Once the module has been imported, you can then choose how you would like to bin
...
+}
+```
+
+There is also the `@Throttle()` decorator which can be used to override the `limit` and `ttl` set in the global module, to give tighter or looser security options. This decorator can be used on a class or a function as well. The order for this decorator does matter, as the arguments are in the order of `limit, ttl`. You have to configure it like this:
+
+```typescript
+// Override default configuration for Rate limiting and duration.
+@Throttle(3, 60)
+@Get()    
+findAll() {
```
micalevisk, line 70:
```suggestion
@Get()
findAll() {
```

micalevisk (approved):

content/security/rate-limiting.md
```diff
@@ -35,13 +35,47 @@ Once the module has been imported, you can then choose how you would like to bin
...
+  // This route will skip rate limiting.
+  @SkipThrottle(false)
+  dontSkip() {
+    return "List users work with Rate limiting"
+  }
+
+  // Rate limiting is applied to this route.
+  doSkip() {
+    return "List users work without Rate limiting"
+  }
```
jmcdo29, line 59: I think the comments are backwards here. `@SkipThrottle(false)` _should_ apply rate-limiting, right?

Tony133: Yes maybe I confused the comments, so if I understand correctly when `@SkipThrottle(false)` is not skipped the rate limiting only when and `@SkipThrottle(true)` right? 

P.s. would you like to send me the change request for this part 😁

jmcdo29: The `@SkipThrottle()` has a default `true` value, hence when you use it you apply the skip metadata. When you use `@SkipThrottle(false)` you are negating a previously applied skip (like wanting to skip for most of a contender but not a certain route).

Tony133: Thanks so much for the explanation, it will be useful to me  😉👍.

P.s. would you like to send me the change request for this part 😁

Tony133: Should I update this part like this:
```typescript
  // Rate limiting is applied to this route.
  @SkipThrottle(false)
  dontSkip() {
    return "List users work with Rate limiting"
  }
  // This route will skip rate limiting.
  doSkip() {
    return "List users work without Rate limiting"
  }
```

right?

Tony133: @jmcdo29, I updated the code snippets. Let me know if I need to make any other changes 👍

jmcdo29 (approved):

PR #2323, opened 2022-05-26
docs(guards): Add missing imports and fix information content

Daniel-Yonkov:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/content/guards.md
+++ b/content/guards.md
@@ -18,7 +18,7 @@ As mentioned, **authorization** is a great use case for Guards because specific 
 @@filename(auth.guard)
 import { Injectable, CanActivate, ExecutionContext } from '@nestjs/common';
 import { Observable } from 'rxjs';
-import { Request } from 'express';
+import type { Request } from 'express';
 import { validateRequest } from './my-custom-authorization'
 
 @Injectable()
@@ -127,7 +127,7 @@ Global guards are used across the whole application, for every controller and ev
 @@filename(app.module)
 import { Module } from '@nestjs/common';
 import { APP_GUARD } from '@nestjs/core';
-import { RolesGuard } from 'roles.guard';
+import { RolesGuard } from './roles.guard';
 
 @Module({
   providers: [
--- a/content/guards.md
+++ b/content/guards.md
@@ -8,7 +8,7 @@ Guards have a **single responsibility**. They determine whether a given request 
 
 But middleware, by its nature, is dumb. It doesn't know which handler will be executed after calling the `next()` function. On the other hand, **Guards** have access to the `ExecutionContext` instance, and thus know exactly what's going to be executed next. They're designed, much like exception filters, pipes, and interceptors, to let you interpose processing logic at exactly the right point in the request/response cycle, and to do so declaratively. This helps keep your code DRY and declarative.
 
-> info **Hint** Guards are executed **after** each middleware, but **before** any interceptor or pipe.
+> info **Hint** Guards are executed **after** all middlewares, but **before** any interceptor or pipe. More on this [here](https://docs.nestjs.com/faq/request-lifecycle#summary).
 
 #### Authorization guard
 
--- a/content/guards.md
+++ b/content/guards.md
@@ -129,6 +129,7 @@ Global guards are used across the whole application, for every controller and ev
 @@filename(app.module)
 import { Module } from '@nestjs/common';
 import { APP_GUARD } from '@nestjs/core';
+import { RolesGuard } from 'roles.guard';
 
 @Module({
   providers: [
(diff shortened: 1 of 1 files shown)


micalevisk (changes requested):
I feel that the line `import { validateRequest } from 'custom-authorization'` could confuse people. What if someone publish a package called `custom-authorization`? 

Also, the snippet below `@@switch` line is the JS version.

content/guards.md
```diff
@@ -125,6 +127,7 @@ Global guards are used across the whole application, for every controller and ev
 @@filename(app.module)
 import { Module } from '@nestjs/common';
 import { APP_GUARD } from '@nestjs/core';
+import { RolesGuard } from 'roles.guard';
```
micalevisk, line 130:
```suggestion
import { RolesGuard } from './roles.guard';
```

I guess?

content/guards.md
```diff
@@ -18,13 +18,15 @@ As mentioned, **authorization** is a great use case for Guards because specific
 @@filename(auth.guard)
 import { Injectable, CanActivate, ExecutionContext } from '@nestjs/common';
 import { Observable } from 'rxjs';
+import { Request } from 'express';
```
micalevisk, line 21:
```suggestion
import type { Request } from 'express';
```

micalevisk (changes requested):

micalevisk:
could you please fix L11 as well?

https://github.com/Daniel-Yonkov/docs.nestjs.com/blob/cc470436fde8189e56a58d30a3d55f4a647af5e5/content/guards.md?#L11

should be

```
> info **Hint** Guards are executed **all** each middleware, but **before** any interceptor or pipe.
```

Daniel-Yonkov:
@micalevisk 
> `> info **Hint** Guards are executed **all** middleware, but **before** any interceptor or pipe.`

I am not sure I understand what is implied in this info - Does this mean that Guards are executed after **all middleware** or before them or are executed based before/after the middleware defined in a dependency chain during runtime?

micalevisk:
@Daniel-Yonkov my bad, should be:

```
> info **Hint** Guards are **after** all middlewares, but **before** any interceptor or pipe.
```

Daniel-Yonkov:
@micalevisk  - Addressed

micalevisk (approved):

Tony133 (approved):

content/guards.md
```diff
@@ -125,6 +127,7 @@ Global guards are used across the whole application, for every controller and ev
 @@filename(app.module)
 import { Module } from '@nestjs/common';
 import { APP_GUARD } from '@nestjs/core';
+import { RolesGuard } from './roles.guard';
```
kamilmysliwiec, line 130: Having `roles.guard.ts` file at the same level as `app.module` does not seem to be an ideal solution, that's why we skipped import (to let developers decide where to put this class).

content/guards.md
```diff
@@ -18,13 +18,15 @@ As mentioned, **authorization** is a great use case for Guards because specific
 @@filename(auth.guard)
 import { Injectable, CanActivate, ExecutionContext } from '@nestjs/common';
 import { Observable } from 'rxjs';
+import type { Request } from 'express';
```
kamilmysliwiec, line 21: We intentionally haven't used express in any example since they are supposed to be platform-agnostic (should work for Fastify as well)

content/guards.md
```diff
@@ -18,13 +18,15 @@ As mentioned, **authorization** is a great use case for Guards because specific
...
+import { validateRequest } from './my-custom-authorization'
```
kamilmysliwiec, line 22: This is quite confusing

content/guards.md
```diff
@@ -8,7 +8,7 @@ Guards have a **single responsibility**. They determine whether a given request
 
 But middleware, by its nature, is dumb. It doesn't know which handler will be executed after calling the `next()` function. On the other hand, **Guards** have access to the `ExecutionContext` instance, and thus know exactly what's going to be executed next. They're designed, much like exception filters, pipes, and interceptors, to let you interpose processing logic at exactly the right point in the request/response cycle, and to do so declaratively. This helps keep your code DRY and declarative.
 
-> info **Hint** Guards are executed **after** each middleware, but **before** any interceptor or pipe.
+> info **Hint** Guards are executed **after** all middlewares, but **before** any interceptor or pipe. More on this [here](https://docs.nestjs.com/faq/request-lifecycle#summary).
```
kamilmysliwiec, line 11: 👍

content/guards.md
Daniel-Yonkov: I understand the reasoning behind giving freedom of where to store the module, but the actual example may introduce a confusion regarding whether the module is injected in some "magical" way since there is no import mentioned. 
Maybe we could go with? 
```suggestion
import { RolesGuard } from './path-to-your-roles.guard';
```

content/guards.md
Daniel-Yonkov: Could you clarify a bit what is confusing? The reasoning behind the import is that it is missing in the example, but it is used to define the authorization for the assigned endpoints. (Line 30)

Issue #2324, opened 2022-05-26
Caching with `CacheManager` and Redis: custom host and port are ignored in `register()`


tukusejssirs:
## I'm submitting a...

- [ ] Regression 
- [ ] Bug report
- [ ] Feature request
- [x] Documentation issue or request (new chapter/page)
- [ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.

## Current behavior

The docs [says](https://github.com/nestjs/docs.nestjs.com/blob/master/content/techniques/caching.md#different-stores) one could define Redis host and port in `register()`, but it does not (it looks like it always tries to connect to `127.0.0.1:6379`).

```typscript
CacheModule.register<RedisClientOptions>({
  store: redisStore,
  socket: {
    host: '192.168.16.2',  // This is ignored and `127.0.0.1` is used instead
    port: 6379
  }
})
```

However, it works with `registerAsync()`:

```typescript

CacheModule.registerAsync({
  useFactory: () => ({
    store: redisStore,
    host: '192.168.16.2',
    port: 6379
  })
})
```

## Expected behavior

Either the docs should be updated (to use `registerAsync()`) or `register()` method should be fixed in order to make it recognised the custom host/port config.

## Minimal reproduction of the problem with instructions

1. Clone [this repo](https://gitlab.com/c10n/test/redis).
2. Start Redis on IP address and/or port different from `127.0.0.1:6379`.
3. Update the IP address and/or port in `AppModule` file.
4. Start the MWE (`npm run start`).
5. It fails with `Error: connect ECONNREFUSED 127.0.0.1:6379`.

## What is the motivation / use case for changing the behavior?

That it does not work as it is stated in the docs. :wink:

#### For Tooling issues:

- Node version: `v18.2.0`

- Platform: Arch Linux x64

#### Others:

See also [this discussion](https://discord.com/channels/520622812742811698/979407192165855252) on Nest Discord channel.

micalevisk:
as discussed on Discord, that snippet is wrong because the package `cache-manager-redis-store` doesn't support the latest version of `redis` (v4) yet: https://github.com/dabroek/node-cache-manager-reds-store/issues/40

thus there's no `socket` option

yepMad:
This problem took up good hours of my day, I was betrayed by Typescript. I made a commit to put this note in the documentation. https://github.com/nestjs/docs.nestjs.com/pull/2333

kamilmysliwiec:
Thank you @yepMad!

PR #2333, opened 2022-06-02
docs(caching): update redis custom host and port

yepMad:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [X] Docs

The documentation takes into account that cache-manager-redis-store supports redis v4, thus leading the reader to believe that they are setting the host and port when this is not read.

Issue Number: #2324

## What is the new behavior?
It informs the actual parameters that are taken into account and explains to the reader why he should download redis major 3 version.

- [ ] Yes
- [X] No

### DIFF
--- a/content/techniques/caching.md
+++ b/content/techniques/caching.md
@@ -220,27 +220,28 @@ class HttpCacheInterceptor extends CacheInterceptor {
 This service takes advantage of [cache-manager](https://github.com/BryanDonovan/node-cache-manager) under the hood. The `cache-manager` package supports a wide-range of useful stores, for example, [Redis store](https://github.com/dabroek/node-cache-manager-redis-store). A full list of supported stores is available [here](https://github.com/BryanDonovan/node-cache-manager#store-engines). To set up the Redis store, simply pass the package together with corresponding options to the `register()` method.
 
 ```typescript
-import type { RedisClientOptions } from 'redis';
+import type { ClientOpts } from 'redis';
 import * as redisStore from 'cache-manager-redis-store';
 import { CacheModule, Module } from '@nestjs/common';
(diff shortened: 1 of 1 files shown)


tukusejssirs:
@yepMad, thanks for the PR!

The Redis installation on [docs.nestjs.com](https://docs.nestjs.com/microservices/redis#installation) already states we should use `redis@3`. Do you think it might be a idea to provide a link to the installation of `redis@3` instead?

It might be a good idea to provide a ‘reverse warning’ in the Redis installation guide to the Redis cache guide and state the underlining problem.

yepMad:
I think it's a good one! As this redis^3 note is only in the microservices part, I didn't pay attention when I was implementing caching. We can also address this note in the caching docs, how about that?

I can do that later if it's ok.

tukusejssirs:
> As this redis^3 note is only in the microservices part, I didn't pay attention when I was implementing caching.

Same here.

Now I think that anything in Nest related to Redis is using `redis@3`, maybe it always uses `cache-manager-redis-store`.

> We can also address this note in the caching docs, how about that?

I agree, but maybe @micalevisk has some other notes. :wink:

tukusejssirs:
> As this redis^3 note is only in the microservices part, I didn't pay attention when I was implementing caching.

Same here.

Now I think that Nest always uses `redis@3`, maybe `cache-manager-redis-store`.

> We can also address this note in the caching docs, how about that?

I agree, but maybe @micalevisk has some other thoughts.

---

PS—Sorry for the comment noise, GitHub Android app has some bugs. 🐛

kamilmysliwiec:
LGTM

PR #2361, opened 2022-06-19
feat: add docs for repl v9 feature

micalevisk:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [x] Docs

## What is the new behavior?

documentation for the feature introduce by this PR: https://github.com/nestjs/nest/pull/9684 that will be available on NestJS v9.

- [ ] Yes
- [x] No


content/recipes/repl.md
```diff
@@ -0,0 +1,102 @@
+### Read-Eval-Print-Loop (REPL)
+
+// TODO: brief description on this feature
```
micalevisk, line 3: idk what to put here

src/app/homepage/menu/menu.component.ts
```diff
@@ -239,6 +239,7 @@ export class MenuComponent implements OnInit {
         { title: 'Prisma', path: '/recipes/prisma' },
         { title: 'Serve static', path: '/recipes/serve-static' },
         { title: 'Commander', path: '/recipes/nest-commander' },
+        { title: 'REPL', path: '/recipes/repl' },
```
kamilmysliwiec, line 242: Let's move REPL to the top of this list 🙌

content/recipes/repl.md
kamilmysliwiec, line 3:
```suggestion
REPL is a simple interactive environment that takes single user inputs, executes them, and returns the result to the user. REPL feature lets you inspect your dependency graph and call methods on your providers (and controllers) directly from your terminal.
```

content/recipes/repl.md
```diff
@@ -0,0 +1,102 @@
...
+bootstrap();
+@@switch
+import { repl } from '@nestjs/core';
+import { AppModule } from './app.module';
+
+async function bootstrap() {
+  await repl(AppModule);
+}
+bootstrap();
+```
```
kamilmysliwiec, line 26: We should probably instruct users to create a separate `repl.ts` file (instead of overriding the `main.ts`).

content/recipes/repl.md
```diff
@@ -0,0 +1,103 @@
...
+
+```text
+> select.help
+Allows navigating through the modules tree, for example, to pull out a specific instance from the selected module.
+Interface: select(token: DynamicModule | ClassRef) => INestApplicationContext
+```
+
+> info **Hint** Those function interfaces are written in [TypeScript function type expression syntax](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-type-expressions).
+
+#### `debug`
```
kamilmysliwiec, line 53: Instead of adding a separate section per method, we could provide a simple table (similar to other chapters in the docs) with a list of available methods + short description. What would be more beneficial to devs (I believe) is an example use-case (a set of commands showcasing how to access a random service + trigger its method from within the terminal + some screenshots/gifs/videos)

micalevisk: I changed to table. The downside is that the interface part won't have syntax highlighting anymore

But I don't know how we could embed images/gifs here :/

kamilmysliwiec: I didn't mean to add gifs in the table (if that's what you're asking about 😄 )

micalevisk: yeah haha

instead of images/gifs what do you think of use https://asciinema.org?

something like

image

I just need to fix the size.

kamilmysliwiec: I've never used it & would love to see it in action :)

content/recipes/repl.md
```diff
@@ -0,0 +1,59 @@
...
+#### Native functions
+
+The built-in NestJS REPL comes with few native functions that are globally available when you start the REPL.
+You can call `help()` to list them out.
+
+If you don't recall what's the signature (ie: expected parameters and return types) of some function, you can call `<function_name>.help`.
+For instance:
+
+```text
+> select.help
```
kamilmysliwiec, line 46: Let's use a different function as an example (`select` will be very rarely used)

Tony133 (approved):

jmcdo29 (approved):
LGTM

content/recipes/repl.md
```diff
@@ -0,0 +1,59 @@
...
+import { repl } from '@nestjs/core';
+import { AppModule } from './app.module';
+
+async function bootstrap() {
+  await repl(AppModule);
+}
+bootstrap();
+```
+
+And then start the app just like the usual:
```
kamilmysliwiec, line 29:
```suggestion
Now in your terminal, start the REPL with the following command:
```

content/recipes/repl.md
```diff
@@ -0,0 +1,64 @@
...
+
+```bash
+$ npm run start --entryFile repl
+```
+
+> info **Hint** `repl` returns a [Node.js REPL server](https://nodejs.org/api/repl.html), thus you can customize it as you wish.
+
+#### Demo
+
+<a href="https://asciinema.org/a/503623" target="_blank"><img src="https://asciinema.org/a/503623.svg" width="636"/></a>
```
micalevisk, line 39: I didn't succeed on embedding the player here but you can try it out: https://asciinema.org/a/503623 & https://asciinema.org/docs/embedding

PR #2375, opened 2022-06-28
Update interceptors.md

davidekete:
The definition of "interceptors" is a bit flawed, using "which" implies that the main requirement for a class to be an interceptor is that it needs to be annotated with the `@injectable` decorator. Which is not entirely true, interceptors can exist without the `@injectable` decorator but it must implement the `NestInterceptor` interface. For clarity, i have changed ",which" to an "and" as it better defines an interceptor.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

A small but significant problem in word choice.

Issue Number: N/A

## What is the new behavior?
a more concise definition of interceptors

- [ ] Yes
- [x] No

### DIFF
--- a/content/interceptors.md
+++ b/content/interceptors.md
@@ -1,6 +1,6 @@
 ### Interceptors
 
-An interceptor is a class annotated with the `@Injectable()` decorator, which implements the `NestInterceptor` interface.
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
LGTM

PR #2417, opened 2022-07-24
docs: change wording of datasource injection scope

mango-martin:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

- [ ] Yes
- [x] No

## Other information
The current wording is a bit confusing. A `project` to me means all the files in the repo. If you think about it, it's obvious that this behavior can't apply to any random file. But by changing the wording a bit I hope it gets clearer as to where the line is actually drawn.

### DIFF
--- a/content/techniques/sql.md
+++ b/content/techniques/sql.md
@@ -61,7 +61,7 @@ The `forRoot()` method supports all the configuration properties exposed by the 
 
 > info **Hint** Learn more about the data source options [here](https://typeorm.io/data-source-options).
 
-Once this is done, the TypeORM `DataSource` and `EntityManager` objects will be available to inject across the entire project (without needing to import any modules), for example:
+Once this is done, the TypeORM `DataSource` and `EntityManager` objects will be available to inject across the entire Nest application's dependency tree (without needing to import any modules), for example:
 
 ```typescript
 @@filename(app.module)


kamilmysliwiec:
IMO this sounds more confusing. cc @jmcdo29 @micalevisk what do you tihnk?

micalevisk:
I think that 'entire project' is fine. Or even 'entire NestJS app'

jmcdo29:
I think the fact that we already mention it will be available to "inject" should imply that this is specifically where DI is possible. I wouldn't expect to inject a value into a class that Nest isn't managing for me.

mango-martin:
Just a suggestion. Feel free to close it, of course.

But if I may I want to add that 
> I think the fact that we already mention it will be available to "inject" should imply that this is specifically where DI is possible. I wouldn't expect to inject a value into a class that Nest isn't managing for me.

This statement in the docs, that you do NOT have to inject it, is partly where the confusion comes from. For anyone who's dealt with Nest for a long time, they read these lines and they know exactly what's going on. But a newcomer might be left confused when they try to use this in a file that they think is "within the project" but then it doesn't work as expected.

mango-martin:
Just a suggestion, feel free to close this, of course.

Issue #2142, opened 2021-12-21
Docs: Serverless FAQ example integration is unrealistic


bchilcott:
## I'm submitting a...

- [ ] Regression 
- [ ] Bug report
- [ ] Feature request
- [x] Documentation issue or request (new chapter/page)
- [ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.

## Current documentation

The documentation at [faq/serverless](https://docs.nestjs.com/faq/serverless#serverless) for 'Example Integration' currently describes how to set up serverless.yml to find an application that has been built in `dist/main.js`. However, using this in practice would require rebuilding every time a change is made, which is unrealistic and *should be* easily improved.

## Expected documentation
This example should reflect a more realistic use case, using the [serverless-plugin-typescript](https://www.npmjs.com/package/serverless-plugin-typescript) and/or [serverless-webpack](https://www.npmjs.com/package/serverless-webpack) to support hot-reloading.

I'll update this and make a PR if/when I figure out how to do this myself, but until then any advice is welcome.

kamilmysliwiec:
Would you like to create a PR for this issue? Contributions are more than welcome!

kamilmysliwiec:
The dev-cycle concern is addressed differently now: faq/serverless.md uses `serverless-offline` for fast local iteration. The specifically proposed plugins are dated advice in the esbuild era. Closing.

PR #2428, opened 2022-07-31
docs(serverless): fixed broken example integration

H4ad:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

Issue Number: #2142 

## What is the new behavior?

Current example integration with serverless is broken and also doesn't provide good default options to reduce package size or have good performance.

So I made some changes:

- Changed the default file from main.ts to lambda.ts, as it makes no sense to build the app just for testing, I recommended keeping main.ts as is and creating lambda.ts with all the necessary code.
- Fixed boilerplate for lambda.ts, the previous example didn't provide a simple fix that could improve cold start by up to a second.
- Be clearer about serverless workflow, no need to build NestJS application every time we want to test.
- Introduce some default extensions and folders to exclude by default to reduce package size.

I introduced NestJS in my company almost 3 years ago migrating from .NET, since then we have deployed serverless APIs using AWS Lambda, we have almost 100 Lambdas in two regions, each lambda being a NestJS API. This PR is a big thanks to the NestJS team and the community for maintaining this great framework that I love so much.

To be honest, I've never used [framework serverless](https://www.serverless.com/framework/docs/) because it didn't work well for my company, I always use built-in tools (like [this one](https://github.com/H4ad/node-modules-packer)) to zip the whole folder and then deploy the zip manually with aws cli. But I the documentation I still kept the serverless framework and added a tip about using `npx serverless package` instead of running `npx serverless deploy`.

- [ ] Yes
- [x] No

## Other information

I also have some questions:

- Shouldn't serverless documentation be a folder inside NestJS docs? The current file for describing serverless documents seems to have a lot of noise because it talks about so many things. It will be better to have a folder and inside it we can cover things like: Optimizations, AWS, Azure, GCP, deep diving into specific libraries when using fastify or express, how to handle asynchronous jobs and many other things.
- Shouldn't the examples in the docs be a project inside the [packages folder](https://github.com/nestjs/nest/tree/master/packages)?


content/faq/serverless.md
```diff
@@ -209,36 +223,53 @@ functions:
...
 > info **Hint** To learn more about the Serverless framework, visit the [official documentation](https://www.serverless.com/framework/docs/).
 
-With this place, we can now navigate to the `main.ts` file and update our bootstrap code with the required boilerplate:
+With that in place, we can now create a file called `lambda.ts` inside `src` and place the necessary boilerplate:
 
 ```typescript
 import { NestFactory } from '@nestjs/core';
-import serverlessExpress from '@vendia/serverless-express';
-import { Callback, Context, Handler } from 'aws-lambda';
+import { default as serverlessExpress } from '@vendia/serverless-express';
```
jmcdo29, line 230: Why not keep the `default` import?

content/faq/serverless.md
```diff
@@ -191,13 +191,27 @@ service: serverless-example
...
+    - '!.idea/**'
+
 provider:
   name: aws
   runtime: nodejs14.x
 
 functions:
   main:
-    handler: dist/main.handler
+    handler: dist/lambda.handler
```
jmcdo29, line 214: Why the change from `main`?

content/faq/serverless.md
```diff
@@ -209,36 +223,53 @@ functions:
...
+  server = server ?? (await delayedFactory);
+
   return server(event, context, callback);
 };
+
 ```
 
 > info **Hint** For creating multiple serverless functions and sharing common modules between them, we recommend using the [CLI Monorepo mode](/cli/monorepo#monorepo-mode).
 
+> info **Hint** It's really recommended to have one file called `setup.ts` to share common configurations to use inside `lambda.ts` and `main.ts`.
```
jmcdo29, line 271: As someone who doesn't work with serverless, is it common to have a `main.ts` for controlled server applications and `lambda.ts` for serverless, or would you normally pick one or the other? Really wondering what the use for both bootstrap files is

content/faq/serverless.md
H4ad: Maybe I configured it wrong in the first tests and it didn't work, now I tested it again and it works fine, I will revert. to just `serverlessExpress`.

content/faq/serverless.md
H4ad: Because now we export the configuration for lambda from `lambda.ts`. The name `main` or `lambda` is based on the file that we want to use to export handler to AWS.

jmcdo29: Right, but I think my question is more towards why are we not using `main.ts` for the lambda handler. Maybe I just didn't understand something elsewhere

content/faq/serverless.md
H4ad: Imagine that you need to configure middlewares, cors, swagger, filters, configure global validation pipes, etc. In my company we use a file to put all this configuration and export as `createApp()`.

In `main.ts` we just run `app.listen()` and inside `lambda.ts` we don't need to call `listen` and we just need to export the handler to AWS.

I didn't put in the documentation the creation of this `setup.ts` file to simplify the process but I can put it instead of leaving a hint.

content/faq/serverless.md
H4ad: During API development in NestJS, we don't use `serverless framework`, we create an application with `createApp` and then call `app.listen`.

I could change the contents of `main.ts`, but that could lead to questions like bchilcott in #2142:

> However, using this in practice would require rebuilding every time a change is made, which is unrealistic and should be easily improved.

So to develop you can run: `npm run start:debug`, then to deploy to AWS Lambda, first test with `npx serverless offline` and then run `npx serverless deploy` to see the changes.

Think of deploying a NestJS API inside AWS Lambda not as deploying a resource or just a method, but the entire API, 10+ resources or something. It's not practical to build and then run `npx serverless offline` and explaining to the user to change commands like `npm run start:debug` just to integrate with AWS Lambda is more troublesome, instead just tell it to create another file.

content/faq/serverless.md
H4ad: Just to give some example, see [this method](https://github.com/H4ad/bird-app/blob/master/backend/src/main.ts#L70), a better organization is to move the `setup` function to another file and then import inside `main.ts`.

jmcdo29: I understand the use of the `setup` method, I use it for running my application and e2e tests in a similar manner. Just not aware of how `serverless offline` acts and thus why there is a need for a `lambda.ts` and a `main.ts`.

H4ad: When you run `serverless offline`, you need to export a function called `handler`. Next, the library will create an http server, then proxy browser requests to the `handler`, then re-proxy the NestJS application with `serverless-express`.

Theoretically, you don't need both, you have both for convenience:

- In `lambda.ts`, we only export the `handler` function.
- In `main.ts`, we just call the `listen` method.

To test locally, we run `npm run start:debug` by calling `main.ts`. To deploy to AWS Lambda, we use `lambda.ts`.
It's not possible to use both because things can get in the way because both options start an http server when running.

Sorry for being redundant, English is not my native language so I'm trying my best to explain, also if you see any English errors in the documentation, please let me know and I'll fix it.

kamilmysliwiec:
These changes seem to make this chapter more confusing & complicated.

content/faq/serverless.md
```diff
@@ -191,13 +191,27 @@ service: serverless-example
...
+    - '!node_modules/**/LICENSE'
+    - '!node_modules/**/*.md'
+    - '!node_modules/**/*.js.map'
+    - '!node_modules/**/*.d.ts.map'
+    - '!node_modules/**/*.d.ts'
+    - '!node_modules/typeorm/browser/**'
+    - '!.webpack/**'
+    - '!src/**'
+    - '!test/**'
+    - '!.idea/**'
```
kamilmysliwiec, line 206: This change doesn't seem to be relevant to NestJS integration. Note: this chapter isn't supposed to serve as a "complete guide to Serverless with Node" but rather showcase how you can use NestJS with serverless, without diving into non-framework-relevant details.

content/faq/serverless.md
kamilmysliwiec: Although I understand the reasoning behind this change, I feel this is material for an article, not something that should be necessarily described in this chapter.

content/faq/serverless.md
```diff
@@ -209,36 +223,53 @@ functions:
...
+// This is a simple trick inspired by this article
+// https://aws.amazon.com/pt/blogs/compute/using-node-js-es-modules-and-top-level-await-in-aws-lambda/
+// This can save you up to one second from cold start.
+const delayedFactory: Promise<Handler> = Promise.resolve()
+  .then(() => bootstrap())
+  .then(app => {
+    server = app;
+
+    return app;
+  });
```
kamilmysliwiec, line 255: Same here https://github.com/nestjs/docs.nestjs.com/pull/2428/files#r934201552

content/faq/serverless.md
kamilmysliwiec: I think the question was - what are the advantages of using `nest start` instead of `serverless offline` (and so maintaining 2 separate files)?

H4ad: With `nest start:debug` every time you change some file it will be reloaded. With `serverless offline`, your changes will only be reflected after you stop the command, run `nest build`, and then run `serverless offline` again.
As bchilcott said, it is unrealistic to have the old workflow as the main default.

content/faq/serverless.md
H4ad: But when we explain how to integrate with AWS Lambda using libraries like `serverless framework` and `@vendia/serverless-express` we already do that. But now, the example is more complete, fixing issues and providing some optimizations.

But if you still think it's not appropriate, I can remove all the node_modules specs and just leave `src`, `test` and `.webpack` because those folders are added by NestJS and the integration, or I can move to the runtime optimization, which do you prefer?

content/faq/serverless.md
H4ad: Should I move this to runtime optimization or just remove it?

content/faq/serverless.md
H4ad: Also, this optimization itself can reduce the package size from 9.9mb to 4.9mb. I know this can be specific but isn't the amount of bundle that could be saved worth leaving as is?

For my own experiences, this optimization is much better than running webpack to bundle the entire nest itself.

Issue #2474, opened 2022-09-29
Documentation : prettier, eslint, vscode ?


cyrilchapon:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

Just coming to Nestjs, it came to my mind that Nest choices around code linting and formating are not clearly documented, and even not clearly stated.

Reading the codebase, 2 npm commands (`lint` / `format`), I can see some packages installed, some config files created, without any explaination. They don't seem to be called by any other command (unless `nest` cli call them magically... ?).

Searching "prettier" / "eslint" / "vscode" in the documentation search bar gives nothing. Searching through github issues leads to corners that don't actually help.

---------

Believe it or not, even after a decade coding in JS / TS, I personally never used prettier in my life (played with all possible eslint / standard and ts flavours though); and apparently Nest made such a choice, that IMHO should be at least stated. From an external point of view, Nest is just not really a framework in the sense of a "library", but rather also a CLI, a set of tools, and opinions. Linter and Formater are part of it.

### Describe the solution you'd like

The need is broad but can be simply expressed :
> As a developer, after installing nest and creating my first project — at least with default configuration, without touching anything else — I want to have a clear idea of the code formatting / styling choices, strategies and tooling involved.

Some examples of what I'd expect to read :

> Nest comes with eslint and prettier preinstalled and configured to work together. Configuration involves packages [eslint, prettier, this-eslint-plugin, etc.]

> Linting and formating can be run through command line, respectively `yarn format` / `yarn lint`

> Configuration is made through those 3 files [eslintrc, .prettierrc, etc.]

> In that default setup, the recommended VSCode extension to work with prettier is [that one].

----------

It could expand to eslint and prettier customization; but this is not my main point here. My main point is really around **default** config explaination.

### Teachability, documentation, adoption, migration strategy

The whole issue is a documentation request, see above.

### What is the motivation / use case for changing the behavior?

The issue is a documentation request, no "behavior change" involved.

jmcdo29:
> The prettier + eslint part is apparently correctly addressed (properly configured to work together); but the implicit "documentation" issue (eg. making vscode working with it) is not.

Can you tell us how it doesn't work with VSCode as it is currently configured? I don't recall ever having an issue with VSC and linting a `nest new` project unless something upstream changed (like Typescript 4.8's AST that broke eslint for a little while)

cyrilchapon:
[...]

PR #2477, opened 2022-09-30
Document code linting and formatting

cyrilchapon:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

Issue Number: #2474

## What is the new behavior?

Linting and formatting documented

- [ ] Yes
- [x] No

### DIFF
--- a/content/first-steps.md
+++ b/content/first-steps.md
@@ -147,6 +147,7 @@ A Nest cli scaffolded project uses
 ##### Code style
 
 With this stack setup, by default we ensure the resulting code follows:
+
 - Typescript eslint [recommended rules](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/recommended.ts)
 - Improved with some [rules in `.eslintrc.js` file](https://github.com/nestjs/typescript-starter/blob/master/.eslintrc.js)
 - Prettier [default options](https://prettier.io/docs/en/options.html)
--- a/content/first-steps.md
+++ b/content/first-steps.md
@@ -115,7 +115,7 @@ This command will watch your files, automatically recompiling and reloading the 
 
 #### Linting / formatting
 
-[Nest CLI](/cli/overview) provides best effort to scaffold a reliable development workflow at scale. Thus, a generated Nest project comes with both a code **formatter** and **linter** preinstalled (respectively [eslint](https://eslint.org/) and [prettier](https://prettier.io/)).
+[Nest CLI](/cli/overview) provides best effort to scaffold a reliable development workflow at scale. Thus, a generated Nest project comes with both a code **linter** and **formatter** preinstalled (respectively [eslint](https://eslint.org/) and [prettier](https://prettier.io/)).
 
 > info **Hint** Not sure about the role of formatters VS linters ? Learn the difference [here](https://prettier.io/docs/en/comparison.html).
 
--- a/content/first-steps.md
+++ b/content/first-steps.md
@@ -113,7 +113,7 @@ $ npm run start:dev
 
 This command will watch your files, automatically recompiling and reloading the server.
 
-#### Linting / formatting
+#### Linting and formatting
 
 [Nest CLI](/cli/overview) provides best effort to scaffold a reliable development workflow at scale. Thus, a generated Nest project comes with both a code **linter** and **formatter** preinstalled (respectively [eslint](https://eslint.org/) and [prettier](https://prettier.io/)).
 
--- a/content/first-steps.md
+++ b/content/first-steps.md
@@ -112,3 +112,49 @@ $ npm run start:dev
 ```
 
 This command will watch your files, automatically recompiling and reloading the server.
+
+#### Linting / formatting
+
+[Nest CLI](/cli/overview) provides best effort to scaffold a reliable development workflow at scale. Thus, a generated Nest project comes with both a code **formatter** and **linter** preinstalled (respectively [eslint](https://eslint.org/) and [prettier](https://prettier.io/)).
+
+> info **Hint** Not sure about the role of formatters VS linters ? Learn the difference [here](https://prettier.io/docs/en/comparison.html).
+
+##### VSCode usage
+To ensure maximum stability and extensibility, we use official [`eslint`](https://www.npmjs.com/package/eslint) and [`prettier`](https://www.npmjs.com/package/prettier) cli packages. This setup allows neat IDE integration with official tooling.
+
+For example with *VSCode*, a Nest project is compatible with [`vscode-eslint`](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) and [`prettier-vscode`](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) by design !
+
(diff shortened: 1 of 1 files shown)


content/first-steps.md
```diff
@@ -112,3 +112,49 @@ $ npm run start:dev
 ```
 
 This command will watch your files, automatically recompiling and reloading the server.
+
+#### Linting / formatting
```
Tony133, line 116:
```suggestion
#### Linting and formatting
```
Otherwise looks good, only the title I would say remove the "/" also to be in line with the titles of the documentation in general.

Tony133 (approved):

content/first-steps.md
```diff
@@ -112,3 +112,49 @@ $ npm run start:dev
...
+
+A Nest cli scaffolded project uses
+- [`eslint-plugin-prettier`](https://github.com/prettier/eslint-plugin-prettier) to remove conflicting rules from `eslint` for use with `prettier`.
+- [`@typescript-eslint/plugin`](https://github.com/typescript-eslint/typescript-eslint/tree/main/packages/eslint-plugin) to enable Typescript linting
+
+> info **Hint** You can learn more about the behavior of `eslint-plugin-prettier` in its [documentation](https://github.com/prettier/eslint-plugin-prettier#recommended-configuration) and discover the actual configuration inside generated [`.eslintrc.js`](https://github.com/nestjs/typescript-starter/blob/master/.eslintrc.js) file
+
+##### Code style
+
+With this stack setup, by default we ensure the resulting code follows :
```
Tony133, line 156:
```suggestion
With this stack setup, by default we ensure the resulting code follows:
```

Tony133 (commented):
one last  little improvement in the text

content/first-steps.md
```diff
@@ -112,3 +112,49 @@ $ npm run start:dev
...
+
+> info **Hint** Not sure about the role of formatters VS linters ? Learn the difference [here](https://prettier.io/docs/en/comparison.html).
+
+##### VSCode usage
+To ensure maximum stability and extensibility, we use official [`eslint`](https://www.npmjs.com/package/eslint) and [`prettier`](https://www.npmjs.com/package/prettier) cli packages. This setup allows neat IDE integration with official tooling.
+
+For example with *VSCode*, a Nest project is compatible with [`vscode-eslint`](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) and [`prettier-vscode`](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) by design !
+
+> warning **Warning** You might be tempted to try [~~`vs-code-prettier-eslint`~~](https://marketplace.visualstudio.com/items?itemName=rvest.vs-code-prettier-eslint). This extension is meant to play with [~~`prettier-eslint`~~](https://github.com/prettier/prettier-eslint) package witch is **not** used inside a Nest project by default.
```
kamilmysliwiec, line 127: Not sure if mentioning any IDE makes sense. I believe it may introduce some confusion, with people thinking that NestJS _recommends_ (or suggests, or simply "integrates best with") using VSCode.

content/first-steps.md
```diff
@@ -112,3 +112,49 @@ $ npm run start:dev
...
+# Format with prettier
+npm run format
+
+# OR
+
+# Lint and autofix with eslint
+yarn lint
+# Format with prettier
+yarn format
+```
```
kamilmysliwiec, line 144: Mentioning only one package manager (yarn or npm) should be sufficient.

content/first-steps.md
cyrilchapon: Good point. Though, despite the fact Nest is not tied to an IDE by any means, its arguably a fact that VSCode is used by a large majority of developers. The part of extensions suggestions ("*For example with VSCode...*" would probably save time for a majority of developers).

I propose
- Renaming paragraph title for agnostism
- Keeping the paragraph "for example"
- Removing the warning which seems a bit too much here

kamilmysliwiec: I still don't think that mentioning IDE-specific extensions in the framework's documentation is a reasonable decision though. If we go down this path, we'd have to create a dedicated page specifically for IDEs and provide docs for the most commonly used IDEs, not only VSCode (which on the other hand brings another "maintenance burden" for us and that's probably not what we want).

cyrilchapon: Alright. Just going to keep the paragraph in an agnostic way, mention neat "IDE integration" (agnostic) and mention the fact we're not using `eslint-prettier` (which might save research time).

All in all, people should understand which extension goes with which package; and Nestjs is providing great native support by design.

cyrilchapon: Does that feel OK like this ?

jmcdo29 (changes requested):
A few newlines to be added and a few slight wording tweaks. Otherwise looks good to me

cyrilchapon:
@jmcdo29 can you please make a bit of squashing ? I did accept suggestions on mobile which wouldn't let me do it in batch

content/first-steps.md
```diff
@@ -112,3 +112,44 @@ $ npm run start:dev
 ```
 
 This command will watch your files, automatically recompiling and reloading the server.
+
+#### Linting and formatting
+
+[Nest CLI](/cli/overview) provides best effort to scaffold a reliable development workflow at scale. Thus, a generated Nest project comes with both a code **linter** and **formatter** preinstalled (respectively [eslint](https://eslint.org/) and [prettier](https://prettier.io/)).
```
micalevisk, line 118:
```suggestion
[Nest CLI](/cli/overview) provides best effort to scaffold a reliable development workflow at scale. Thus, a generated Nest project comes with both a code **linter** and **formatter** preinstalled (respectively [ESLint](https://eslint.org/) and [Prettier](https://prettier.io/)). Do note that you can use any other tool for such tasks, and they are not maintained by the NestJS core team.
```

I guess we should mention somehow that ESLint & Prettier are not related with NestJS directly, otherwise beginners on Node might think that they are.

cyrilchapon:
Guys, sorry to set such a cat among the pigeons, but I don't understand why we are taking so many precautions here.

As an example, here is the [documentation about testing](https://docs.nestjs.com/fundamentals/testing) which is (conceptually) in the same field. Jest and Supertest are mentioned and linked, there are code examples, there is the word "support integration" clearly typed down; and no warning about "Nest not maintaining them" of some sort, appart mentioning the fact Nest "is agnostic" about testing frameworks — which I believe I replicated in the linting and formating part.

- You guys apparently think people will be mislead and think eslint and prettier are actually maintained by Nest core team.
- What I rather believe is people will expect eslint and prettier **integration configuration** to be maintained by Nest core team.

(I could be wrong here, feel free to disabuse me. I mean my analogy could be an actual bad one, and the current testing documentation could actually be a problem leading to parasitic issues of people thinking Jest is maintained by NestJS. Is it ?)

So one of two things
- Either we fear the bad thing : to actually maintain the eslint + prettier integration and configuration here. In that case I think we have an issue, since that configuration is a true part of NestJS product, and **should** be maintained by it. I mean, this is not some sort of beta plugin or community addon; it's a core integration, shipped with Nest, with tools chosen by Nest, pre-configured by Nest.
- Either I completely missed the point and wrote that piece of doc entirely in a wrong way that presents eslint and prettier as internal tools. Feel free to tell me then, my English is far from perfect, and I can rewrite this.

If none of the above, I don't understand while more precautions should be taken than other parts of documentation, like testing with Jest (which actually has both less popularity than eslint plus an strong name ambiguousness with NestJS...). If noone thinks Jest is maintained by Nest; why would someone think eslint is maintained by Nest ?

micalevisk (approved):

PR #2487, opened 2022-10-03
fix: specify max cache-manager version

botika:
I have refactored cache manager and in version 5 it is not compatible with the current implementation in NestJS. 
https://github.com/node-cache-manager/node-cache-manager/pull/203


jmcdo29 (approved):

Tony133 (approved):

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest/pull/10346

PR #2492, opened 2022-10-08
Make it clear that TTL is in milliseconds

ahernandez2-rbi:
The way it was documented can lead people to think that TTL is in seconds

See https://github.com/nestjs/nest/pull/10380


micalevisk:
We should mention that that it is in seconds in `cache-manager` v5+

PR #2493, opened 2022-10-14
docs(caching): fix github link to point to the moved repo location

MitchellCash:
The `node-cache-manager` repo was moved to a new parent account on GitHub.

### DIFF
--- a/content/techniques/caching.md
+++ b/content/techniques/caching.md
@@ -4,11 +4,10 @@ Caching is a great and simple **technique** that helps improve your app's perfor
 
 #### Installation
 
-First install [required packages](https://github.com/node-cache-manager/node-cache-manager) (and its types for TypeScript users):
+First install [required packages](https://github.com/node-cache-manager/node-cache-manager):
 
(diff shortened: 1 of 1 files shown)


Tony133 (approved):

MitchellCash:
As an aside, as of `v5` of `node-cache-manager` it now provides its own type definitions so we no longer need the line: 

```sh
npm install -D @types/cache-manager
```

However, `v5` is only 15 days old and I haven't tested if the new major version is compatible with Nest so I didn't remove this line. If someone more knowledgable than I can confirm Nest works with `v5` than I can also go ahead and remove that additional dependency requirement.

micalevisk:
@MitchellCash as we omit the version, we're assuming that people will now install the latest one, so we can drop that line

MitchellCash:
Thanks @micalevisk, new commit removes the line

micalevisk (approved):

kamilmysliwiec:
lgtm

PR #2504, opened 2022-10-20
docs(Introduction): Corrected a grammatical error

Sirjoseph94:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

`is built with and fully supports [TypeScript](http://www.typescriptlang.org/)...`

Issue Number: N/A

## What is the new behavior?
`It is built with and fully supports [TypeScript](http://www.typescriptlang.org/)...`

- [ ] Yes
- [x] No

## Other information
A minor grammatical fix on the docs introduction


jmcdo29:
The current grammar is fine. If we were to add the `it` where you've proposed, we'd need to add it to the lat thing as well. This is basically saying saying `It does this, that, and another thing`. Your proposed change would need to be `It does this, it is that, and it does this other thing`

Sirjoseph94:
Okay, I see what you're saying. When I initially read it, it was a little unclear. That's why I raised the PR.

Thanks for the feedback

PR #2506, opened 2022-10-22
Correction on exceptions filter for httpAdapterHost

WayneJr:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes, no api changes)
- [x] Build related changes
- [ ] Docs

At present the docs states that the variable to be passed into the AllExceptionsFilter should be an httpAdapter, but the constructor requests for an HttpAdapterHost and this breaks the typescript.

Issue Number: N/A

## What is the new behavior?
It compiles successfully after the change to an httpAdapterHost variable.

- [ ] Yes
- [x] No

### DIFF
--- a/content/exception-filters.md
+++ b/content/exception-filters.md
@@ -364,8 +364,8 @@ The first method is to inject the `HttpAdapter` reference when instantiating the
 async function bootstrap() {
   const app = await NestFactory.create(AppModule);
 
-  const { httpAdapter } = app.get(HttpAdapterHost);
-  app.useGlobalFilters(new AllExceptionsFilter(httpAdapter));
+  const httpAdapterHost = app.get(HttpAdapterHost);
+  app.useGlobalFilters(new AllExceptionsFilter(httpAdapterHost));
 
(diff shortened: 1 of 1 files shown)


content/exception-filters.md
```diff
@@ -364,8 +364,8 @@ The first method is to inject the `HttpAdapter` reference when instantiating the
 async function bootstrap() {
   const app = await NestFactory.create(AppModule);
 
-  const { httpAdapter } = app.get(HttpAdapterHost);
-  app.useGlobalFilters(new AllExceptionsFilter(httpAdapter));
```
jmcdo29, line 368: This `AllExceptionsFilter` is not the same as the one up on line 300. This one `extends BaseExceptionFilter` which takes in an `HttpAdapter` instead of an `HttpAdapterHost`. The docs are correct

WayneJr: Alright, that makes sense but it's not this clear in the docs, it almost looks like it's just a continuation.

PR #2507, opened 2022-10-23
docs(recipes): add async local storage recipe

Papooch:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [X] Docs

There are no docs on using AsyncLocalStorage with NestJS.

## What is the new behavior?
There are some.

- [ ] Yes
- [X] No

## Other information
Since most parts of the [AsyncLocalStorage](https://nodejs.org/api/async_context.html#asynchronous-context-tracking) Node.JS API are already stable now, and since other frameworks are already including first-class support for it ([Fastify](https://github.com/fastify/fastify-request-context), [Adonis](https://docs.adonisjs.com/guides/context#access-http-context-from-anywhere)), as per @jmcdo29's suggestion, I added a recipe on how to implement it in NestJS and also added a sneak peek of my library that's meant to provide a nice DX on top of `AsyncLocalStorage`.


content/recipes/async-local-storage.md
```diff
@@ -0,0 +1,228 @@
...
+
+3. To get strong typing of the store values managed by the `ClsService`, we can use an optional type parameter (`ClsService<MyStoreType>`), or use application-wide typescript module augmentation.
+
+```ts
+declare module `nestjs-cls` {
+  interface ClsStore {
+    userId: number
+  }
+}
+```
```
jmcdo29, line 169: nit: If there's a way to use a custom interface over module augmentation, I'd prefer we show that.

jmcdo29 (approved):
Overall, looks great! I like the introduction to Async Local Storage and then going into your package with it. We'll see if Kamil has anything else to mention

content/recipes/async-local-storage.md
```diff
@@ -0,0 +1,228 @@
...
+
+The main idea of Async Local Storage is that we can _wrap_ some function call with the `AsyncLocalStorage#run` call. All code that is invoked within the wrapped call gets access to the same `store`, which will be unique to each call chain.
+
+In the context of NestJS, that means if we can find a place within the request's lifecycle where we can wrap the rest of the request's code, we will be able to access and modify state visible only to that request, effectively eliminating the need for REQUEST scoped providers and their limitations.
+
+#### Custom implementation
+
+NestJS itself does not provide any built-in abstraction for `AsyncLocalStorage`, so let's walk through how we could implement it ourselves for the simplest HTTP case to get a better understanding of the whole concept:
+
+> info **info** For a ready-made [dedicated package](#nestjs-cls), continue reading below.
```
jmcdo29, line 13: I'm not sure if this anchor works as expected when used as a website. Might be the only thing to double check. The preview linked me back to the introduction page

content/recipes/async-local-storage.md
```diff
@@ -0,0 +1,227 @@
...
+@Module({
+  providers: [
+    {
+      provide: AsyncLocalStorage,
+      useValue: asyncLocalStorage,
+    },
+  ],
+  exports: [AsyncLocalStorage],
+})
+export class AlsModule {}
```
micalevisk, line 34:
```suggestion
@@filename(als.module)
import { AsyncLocalStorage } from 'async_hooks';
import { Module } from '@nestjs/core';

export const asyncLocalStorage = new AsyncLocalStorage();

@Module({
  providers: [
    {
      provide: AsyncLocalStorage,
      useValue: asyncLocalStorage,
    },
  ],
  exports: [AsyncLocalStorage],
})
export class AlsModule {}
```

content/recipes/async-local-storage.md
```diff
@@ -0,0 +1,227 @@
...
+
+NestJS itself does not provide any built-in abstraction for `AsyncLocalStorage`, so let's walk through how we could implement it ourselves for the simplest HTTP case to get a better understanding of the whole concept:
+
+> info **info** For a ready-made [dedicated package](recipes/async-local-storage#nestjs-cls), continue reading below.
+
+1. First, create a new instance of the `AsyncLocalStorage` in some shared source file. Since we're using NestJS, let's also turn it into a module with a custom provider.
+
+```ts
+/** als.module.ts */
+
```
micalevisk, line 19:
```suggestion
@@filename(als.module)
```

content/recipes/async-local-storage.md
```diff
@@ -0,0 +1,227 @@
...
+  exports: [AsyncLocalStorage],
+})
+export class AlsModule {}
+```
+
+2. We're only concerned with HTTP, so let's use a middleware to wrap the `next` function with `AsyncLocalStorage#run`. Since a middleware is the first thing that the request hits, this will make the `store` available in all enhancers and the rest of the system.
+
+```ts
+/** main.ts */
+
```
micalevisk, line 41:
```suggestion
@@filename(main)
```

content/recipes/async-local-storage.md
```diff
@@ -0,0 +1,227 @@
...
+}
+
+bootstrap();
+```
+
+3. Now, anywhere within the lifecycle of a request, we can access the local store instance.
+
+```ts
+/** cat.service.ts */
+
```
micalevisk, line 73:
```suggestion
@@filename(cat.service)
```

content/recipes/async-local-storage.md
```diff
@@ -0,0 +1,227 @@
...
+#### Usage
+
+A similar functionality as described [above](#custom-implementation) can be implemented using `nestjs-cls` as follows:
+
+1. Import the `ClsModule` in the root module.
+
+```ts
+/** app.module.ts */
+
```
micalevisk, line 118:
```suggestion
@@filename(app.module)
```

content/recipes/async-local-storage.md
```diff
@@ -0,0 +1,227 @@
...
+  controllers: [CatController],
+})
+export class AppModule {}
+```
+
+2. And then can use the `ClsService` to access the store values.
+
+```ts
+/** cat.service.ts */
+
```
micalevisk, line 145:
```suggestion
@@filename(cat.service)
```

content/recipes/async-local-storage.md
```diff
@@ -0,0 +1,227 @@
...
+```
+
+2. We're only concerned with HTTP, so let's use a middleware to wrap the `next` function with `AsyncLocalStorage#run`. Since a middleware is the first thing that the request hits, this will make the `store` available in all enhancers and the rest of the system.
+
+```ts
+/** main.ts */
+
+import { NestFactory } from '@nestjs/core';
+import { AppModule } from './app.module';
+import { asyncLocalStorage } from './als.setup.ts';
```
micalevisk, line 44: why we have to import the instance here?

also, 
```suggestion
import { asyncLocalStorage } from './als.module.ts';
```

content/recipes/async-local-storage.md
```diff
@@ -0,0 +1,227 @@
...
+
+1. First, create a new instance of the `AsyncLocalStorage` in some shared source file. Since we're using NestJS, let's also turn it into a module with a custom provider.
+
+```ts
+/** als.module.ts */
+
+import { AsyncLocalStorage } from 'async_hooks';
+import { Module } from '@nestjs/core';
+
+export const asyncLocalStorage = new AsyncLocalStorage();
```
micalevisk, line 23: not sure if the instance will be needed somewhere

If not, then we can move this to L29 instead, turn it into an in-line version

content/recipes/async-local-storage.md
Papooch: Oh, I overlooked that when re-facring the example :)

content/recipes/async-local-storage.md
Papooch: I thought about that also. But we can use the instance outside of the DI context if needed. But it's an advanced use case (and possibly an anti pattern), so yeah, let's not expose it in the example.

content/recipes/async-local-storage.md
```diff
@@ -0,0 +1,227 @@
...
+  const als = app.get(AsyncLocalStorage);
+
+  // Here we can bind the middleware to all routes,
+  app.use((req, res, next) => {
+    // populate the store with some default values
+    // based on the request,
+    const store = {
+      userId: req.headers['x-user-id'],
+    };
+    // and and pass the "next" function as callback
```
micalevisk, line 60:
```suggestion
    // and pass the "next" function as callback
```

content/recipes/async-local-storage.md
```diff
@@ -0,0 +1,227 @@
...
+
+  describe('getCatForUser', () => {
+    it('retrieves cat based on user id', async () => {
+      const expectedUserId = 42
+      mockCatRepository.getForUser.mockImplementationOnce(
+        (id) => ({ userId: id })
+      )
+
+      // Wrap the test call the `runWith` method
+      // in which we can pass hand-crafted store values.
```
micalevisk, line 213:
```suggestion
      // Wrap the test call in the `runWith` method
      // in which we can pass hand-crafted store values.
```

I guess?

content/recipes/async-local-storage.md
Papooch: This causes the build to fail with this error
image
Do you know what might be the cause?

Papooch: It's also strange, because the same syntax is used in `cqrs.component.ts`, which is part of the same `recipes.module.ts` as `async-local-storage.component.ts`

content/recipes/async-local-storage.md
Papooch: I guess I could also put `providers: [AsyncLocalStorage]` directly and NestJS would automatically create an instace, but I wanted to be explicit here.

content/recipes/async-local-storage.md
micalevisk: What if we change `ts` to `typescript`

Papooch: No change. I even tried copy-and-pasting an entire block from the `cqrs.ts` file and the error is still there. :disappointed:

Papooch: I even tried putting the `ExtensionPipe` directly in the `RecipesModule`, but that only caused more errors. That's where my angular debugging proficiency ends. @jmcdo29, would you have any idea what might be wrong?

jmcdo29: Huh, I'll have to work with this locally and try to figure out what that's about

content/recipes/async-local-storage.md
```diff
@@ -0,0 +1,225 @@
+### Async Local Storage
+
+`AsyncLocalStorage` is a [Node.js API](https://nodejs.org/api/async_context.html#async_context_class_asynclocalstorage) (based on the `async_hooks` API) that provides an alternative way of propagating local state through the application without the need to explicitly pass it as a function parameter. It is similar to a thread-local storage in other languages.
+
+The main idea of Async Local Storage is that we can _wrap_ some function call with the `AsyncLocalStorage#run` call. All code that is invoked within the wrapped call gets access to the same `store`, which will be unique to each call chain.
+
+In the context of NestJS, that means if we can find a place within the request's lifecycle where we can wrap the rest of the request's code, we will be able to access and modify state visible only to that request, effectively eliminating the need for REQUEST scoped providers and their limitations.
```
kamilmysliwiec, line 7: REQUEST-scoped providers have some other great use cases, like constructing providers tied to specific requests so we might need to mention here that these two approaches are not really interchangeable.

content/recipes/async-local-storage.md
```diff
@@ -0,0 +1,225 @@
...
+    // populate the store with some default values
+    // based on the request,
+    const store = {
+      userId: req.headers['x-user-id'],
+    };
+    // and pass the "next" function as callback
+    // to the "als.run" method together with the store.
+    als.run(store, () => next());
+  });
+}
```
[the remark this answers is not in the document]
svanda-navisys: I wanted to show a solution with the least amount of code possible, so it's as on-point as possible, but you're right, we want to enforce best practices here. I'll move it to `AppModule#configure` method as creating a class middleware seems like an overkill in this scenario.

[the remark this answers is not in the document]
Papooch: I wanted to show a solution with the least amount of code possible, so it's as on-point as possible, but you're right, we want to enforce best practices here. I'll move it to `AppModule#configure` method as creating a class middleware seems like an overkill in this scenario.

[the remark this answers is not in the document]
kamilmysliwiec: Sounds great 🙌

content/recipes/async-local-storage.md
Papooch: Constructing providers per request is also possible with the Proxy Providers feature of `nestjs-cls`, but yeah, I'll try to reword it, so it's not so harsh on REQUEST-scoped providers.

Papooch: Sure, but REQUEST-scoped providers have limitations and simply cannot be used in certain cases, and that's the whole point of using AsyncLocalStorage, which doesn't require consuming providers to also become request-scoped.

With proxy providers, I can create the same "request" scoped provider for a CRON job as well as for a HTTP request and consume it in my services using the same API (any my services don't need to be aware of the notion of a request, hence separating concerns). This is not re-inventing a wheel per-se, it's just offering a different set of wheels for a similar functionality.

Don't get me wrong, I'm not trying to undermine the features of the framework, but I'll make a bold statement that ALS can completely replace REQUEST-scoped providers. And when (if?) ALS matures enough to be entirely stable and introduce neglibigle runtime overhead, we might even go as far as to reimplement actual REQUEST-scoped providers on top of it in NestJS itself.

kamilmysliwiec: > With proxy providers, I can create the same "request" scoped provider for a CRON job as well as for a HTTP request or Kafka message and consume it in my services using the same API (any my services don't need to be aware of the notion of doferent transports, hence separating concerns). This is not re-inventing a wheel per-se, it's just offering a different set of wheels for a similar functionality.

I don't see how this feature differs from the built-in request-scoped providers. This should be already feasible unless I misunderstood what you're trying to accomplish.

> But I'll make a bold statement that ALS can completely replace REQUEST-scoped providers.

I'm not saying that it _can't completely replace them_. I'm just saying that _it shouldn't_ because it _makes no sense_.

kamilmysliwiec: Also, just to make sure we're on the same page. I'm not saying that using ALS is wrong by any means. ALS is great and it has some nice use cases. What I'm trying to emphasize is - using ALS to construct request-scoped providers doesn't seem to be the cleanest and optimal solution when you can already accomplish what you need by utilizing custom DI sub-trees (if you need dynamic instantiation, transport strategy agnostic) or with simple request-scoped providers (if the static hierarchy is fine) 

ALS and request-scoped providers are not interchangeable. Ideally, we could (and would want, whenever ALS becomes stable and mature enough) leverage both depending on what we want to accomplish (possibly even at the same time).

svanda-navisys: > I don't see how this feature differs from the built-in request-scoped providers. This should be already feasible unless I misunderstood what you're trying to accomplish.

Imagine I have a multi tenant app, which has multiple entry points, but they all eventually use the same set of shared services. There are the following requirements:
1) limit access only to records based on the `tenantId` of the request,
2) restrict certain functionality based on the `userRole of the user making that request`

I could pass these variables down to every method or event, but you understand that it would be an absolute nightmare to maintain and re-factor.

So, using `nestjs-cls`, I create a Proxy Provider called "RequestContext" that has 2 properties - `userRole` and `tenantId`. 

In my app, I have HTTP endpoints, Kafka consumers, CRON controllers. Each of which has a different way of extracting the context from the message and populating the ALS. I can feed the ALS (in three different ways) with it and construct the `RequestContext` object based on the values in ALS. The rest of the app only depends on the `RequestContext` provider.

There's no REQUEST in Kafka, there's only CONTEXT - so I don't see how I could create a shared provider based on both.

Additionally, last I checked, there was no way to use services that depend on the REQUEST in a CRON handler.

So, how would you do something like this using built-in REQUEST-scoped providers?

(Sorry, I feel like I'm diverging significantly from the original topic 😅, but I'd love to have a discussion about application design sometime)

kamilmysliwiec: > There's no REQUEST in Kafka, there's only CONTEXT - so I don't see how I could create a shared provider based on both.

`CONTEXT` is just an alias for `REQUEST`. You could create a `RequestContext` factory provider that injects `REQUEST` and extracts "the context data" based on the transport layer.

> Additionally, last I checked, there was no way to use services that depend on the REQUEST in a CRON handler.

AFAIR it's feasible with Bull crons, I can't recall if we support it out-of-the-box with nestjs/schedule. Still, you can dynamically instantiate request-scoped providers from statically-scoped providers using custom DI sub-trees (and registering a custom, platform-agnostic RequestContext object for that subtree).

content/recipes/async-local-storage.md
jmcdo29: You need to add `ASyncLocalStorageComponents` to the `RecipeModule`'s `declarations`

```diff                                                   
      ServeStaticComponent,                                                              
      NestCommanderComponent,                                                            
      ReplComponent,                                                                     
+     AsyncLocalStorageComponent,                                                        
    ],                                                                                   
  })                                                                                     
- export class RecipesModule {}                                                          
+ export class RecipesModule { }                                                                                                                                                
```

Papooch:
I'm sorry to bother, is there anything else that prevents this PR from getting merged in the current state?

micalevisk:
@Papooch can you click on 'Resolve conversation' button on comments already done?

Papooch:
@micalevisk My uderstanding was that it's the original commenters who has to resolve the conversation once they're satisfied with the change, but sure I can do that.

micalevisk:
@Papooch nvm you're correct lol

Papooch:
@kamilmysliwiec Sorry for the ping, do you have any more comments on this?

Papooch:
Just wanted to point out there's already a mention of `AsyncLocalStorage` in the [MikroORM recipe](https://docs.nestjs.com/recipes/mikroorm#using-asynclocalstorage-for-request-context), should I add a cross-reference?

kamilmysliwiec:
lgtm

PR #2530, opened 2022-11-12
docs(caching): TTL in ms as from cache-manager@^5

alexismch:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

In caching doc, the TTL is defined in seconds.

Issue Number: N/A

## What is the new behavior?
As from cache-manager@^5, the TTL is defined in milliseconds.

- [ ] Yes
- [x] No

### DIFF
--- a/content/techniques/caching.md
+++ b/content/techniques/caching.md
@@ -51,7 +51,7 @@ await this.cacheManager.set('key', 'value');
 
 The default expiration time of the cache is 5 seconds.
 
-You can manually specify a TTL (expiration time in seconds) for this specific key, as follows:
+You can manually specify a TTL (expiration time in seconds, milliseconds as from cache-manager@^5) for this specific key, as follows:
 
 ```typescript
 await this.cacheManager.set('key', 'value', 1000);
@@ -121,7 +121,7 @@ All cached data has its own expiration time ([TTL](https://en.wikipedia.org/wiki
 
 ```typescript
 CacheModule.register({
-  ttl: 5, // seconds
+  ttl: 5, // seconds, milliseconds as from cache-manager@^5
   max: 10, // maximum number of items in cache
 });
 ```


micalevisk:
in the latest version of `@nestjs/common`, the `ttl` opt is still in seconds regardless of the version of `cahe-manager`. See: https://github.com/nestjs/nest/blob/8e3af065bb9abbecc91b4e6068c8e205ef79d165/packages/common/cache/cache.providers.ts#L25-L32 (if it is v5+ that option is multiplied by 1000)

alexismch:
> in the latest version of `@nestjs/common`, the `ttl` opt is still in seconds regardless of the version of `cahe-manager`. See: https://github.com/nestjs/nest/blob/8e3af065bb9abbecc91b4e6068c8e205ef79d165/packages/common/cache/cache.providers.ts#L25-L32 (if it is v5+ that option is multiplied by 1000)

@micalevisk As seen in the code you linked, the TTL is multiplied by 1000 only for the default option, not for the passed one.
I refer as well to the CacheManagerOptions (c.f. https://github.com/nestjs/nest/blob/8e3af065bb9abbecc91b4e6068c8e205ef79d165/packages/common/cache/interfaces/cache-manager.interface.ts#L72-L78) comment that states the same words as my patch.

alexismch:
@micalevisk can you please review my last comment and maybe reconsider my PR? Thanks

micalevisk:
@alexismch yeah, you were correct. Can't you reopen this?

alexismch:
@micalevisk thank you for your return. Sadly it seems not, I think only @kamilmysliwiec can do so

micalevisk:
tbh I'd prefer if we change that to always be in seconds. But then it would introduces breaking changes :/ that's why we should document it here

PR #2559, opened 2022-12-12
docs(security): add `implements AbstractStrategy` to improve DX

### LINKED ISSUE #2455: Implement AbstractStrategy in https://docs.nestjs.com/security/authentication#implementing-passport-local ?
## I'm submitting a...

- [ ] Regression 
- [ ] Bug report
- [ ] Feature request
- [x] Documentation issue or request (new chapter/page)
- [ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.

## Current behavior

As https://docs.nestjs.com/security/authentication#implementing-passport-strategies and 
 https://docs.nestjs.com/security/authentication#implementing-passport-local says, you need to implement a `validate(args[])` function yourself for a subclass of PassportStrategy, but the IDE will consider `validated(args[])` unused.

image

## Expected behavior

Maybe we should encourage users to implement `AbstractStrategy` from `@nestjs/passport` to get better editing experience.

image

## Minimal reproduction of the problem with instructions

As the screenshot.

For better editing experience, and IDE will promot user to implement `validate(args[])`.

#### For Tooling issues:

- Node version: not related

- Platform: not related

#### Others:

It will be better if we could let PassportStrategy implement AbstractStrategy.

micalevisk:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

Issue Number: closes #2455

- [ ] Yes
- [x] No

### DIFF
--- a/content/security/authentication.md
+++ b/content/security/authentication.md
@@ -205,12 +205,12 @@ Now we can implement our Passport **local authentication strategy**. Create a fi
 ```typescript
 @@filename(auth/local.strategy)
 import { Strategy } from 'passport-local';
-import { PassportStrategy } from '@nestjs/passport';
+import { PassportStrategy, AbstractStrategy } from '@nestjs/passport';
 import { Injectable, UnauthorizedException } from '@nestjs/common';
 import { AuthService } from './auth.service';
 
 @Injectable()
-export class LocalStrategy extends PassportStrategy(Strategy) {
+export class LocalStrategy extends PassportStrategy(Strategy) implements AbstractStrategy {
   constructor(private authService: AuthService) {
     super();
   }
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
Are we sure that this actually improves DX? With this change, we'd have an additional implements instruction even though it's not actually needed to make things work (+ `extends PassportStrategy` already enforces you to provide the validate method).

DevDengChao:
> `extends PassportStrategy` already enforces you to provide the validate method

@kamilmysliwiec Sorry but this is not ture. 
Extending `PassportStrategy` didn't enforced the developer to implement `validate()` before running the application as the `PassportStrategy` is defined as a function not an interface.
The `passport.strategy.d.ts` didn't metioned it too.

kamilmysliwiec:
@DevDengChao We indeed might have changed that for some reason 🤔 Good catch! We should probably just update the `PassportStrategy` function then https://github.com/nestjs/passport/blob/master/lib/passport/passport.strategy.ts#LL8

DevDengChao:
@micalevisk It's closed without more comments, dose it means this will not be implemented ? or where should I trace this then ? 👀

micalevisk:
> @micalevisk It's closed without more comments, dose it means this will not be implemented ? or where should I trace this then ? 👀

I agree with Kamil's last comment. That's why I close this.

DevDengChao:
> > @micalevisk It's closed without more comments, dose it means this will not be implemented ? or where should I trace this then ? 👀
> 
> I agree with Kamil's last comment. That's why I close this.

Is there a new issue/pr that I could subscribe ?

micalevisk:
There is not. Feel free to open a new PR

PR #2565, opened 2022-12-18
docs(sql): remove extra `+` added in sql section

dammy001:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] Docs

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/content/techniques/sql.md
+++ b/content/techniques/sql.md
@@ -632,7 +632,7 @@ TypeOrmModule.forRootAsync({
   useFactory: (configService: ConfigService) => ({
     type: 'mysql',
     host: configService.get('HOST'),
-    port: +configService.get('PORT'),
+    port: configService.get('PORT'),
     username: configService.get('USERNAME'),
     password: configService.get('PASSWORD'),
     database: configService.get('DATABASE'),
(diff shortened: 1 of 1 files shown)


jmcdo29:
The `+` is there to coerce the value to a `Number`instead of being interpreted as a `String`. Does TypeORM and Sequelize both accept strings for the port?

dammy001:
> The `+` is there to coerce the value to a `Number`instead of being interpreted as a `String`. Does TypeORM and Sequelize both accept strings for the port?

Yes, TypeORM accepts string so I guess Sequelize should do so as well.

jmcdo29:
> > The `+` is there to coerce the value to a `Number`instead of being interpreted as a `String`. Does TypeORM and Sequelize both accept strings for the port?
> 
> Yes, TypeORM accepts string so I guess Sequelize should do so as well.

[Looks like they don't accept a string](https://sequelize.org/api/v6/class/src/sequelize.js~sequelize#instance-constructor-constructor)

Even if in the code it works, I'd rather we keep this explicitly coerced to a number via JS standards rather than hope that it works without and have people run to us that the docs are incorrect

PR #2577, opened 2023-01-04
docs(authentitcation): update the example secret

lmarqs:
Update de secret value used as example in order to avoid programmers to use it on production code

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

I've found a few companies using the example code without modifying the secret value, regardless the warning on the page.
I think that it is a honest mistake, because the current value seems like a enum value.

Issue Number: N/A

## What is the new behavior?

The goal of this modification is to use a more semantic value, informing to the the programmer that the secret value must be changed and kept safe.

- [ ] Yes
- [x] No

### DIFF
--- a/content/security/authentication.md
+++ b/content/security/authentication.md
@@ -465,11 +465,11 @@ First, create `constants.ts` in the `auth` folder, and add the following code:
 ```typescript
 @@filename(auth/constants)
 export const jwtConstants = {
-  secret: 'secretKey',
+  secret: 'DO NOT USE THIS VALUE. INSTEAD, CREATE A COMPLEX SECRET AND KEEP IT SAFE OUTSIDE OF THE SOURCE CODE.',
 };
 @@switch
 export const jwtConstants = {
-  secret: 'secretKey',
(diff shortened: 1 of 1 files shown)


content/security/authentication.md
```diff
@@ -465,11 +465,11 @@ First, create `constants.ts` in the `auth` folder, and add the following code:
 ```typescript
 @@filename(auth/constants)
 export const jwtConstants = {
-  secret: 'secretKey',
+  secret: 'DO NOT USE THIS VALUE. INSTEAD, CREATE A COMPLEX SECRET AND KEEP IT SAFE.',
```
micalevisk, line 468: I guess we could even have "OUTSIDE OF THE SOURCE CODE"

lmarqs: Good call. I've included it.

lmarqs:
@micalevisk I miss the `OF` in your suggestion.
Could you please review it one more time?

micalevisk (approved):

PR #2580, opened 2023-01-05
docs(queues): add hint to start consuming from queues

kikar:
## PR Checklist

## PR Type

- [x] Docs

No explanation on how Nest starts consuming from queues.

## What is the new behavior?
Explains how Nest starts consuming from queues.

- [x] No

### DIFF
--- a/content/techniques/queues.md
+++ b/content/techniques/queues.md
@@ -207,6 +207,16 @@ export class AudioConsumer {}
 
 > info **Hint** Consumers must be registered as `providers` so the `@nestjs/bull` package can pick them up.
 
+Once registered as providers, the consumer will start consuming from the queue only if the app is either listening or the `init` hook has been called:
+
+```typescript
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
Hmm this should be obvious after reading other chapters in the documentation. I'll see how we could make that more straightforward but adding such a hint in the queues related page is a slippery slope (we might have ended up adding it for every integration)

kikar:
> Hmm this should be obvious after reading other chapters in the documentation. I'll see how we could make that more straightforward but adding such a hint in the queues related page is a slippery slope (we might have ended up adding it for every integration)

@kamilmysliwiec I've been using Nest for over 6 years, and still took me over 30 mins to debug the issue cause I never used this library before.
Also, just searched for the word `init` in the docs, and I don't see any reference where it's explained. Cause you please point me to it?

PR #2583, opened 2023-01-05
docs(overview): add warning for use provider app_guard

Tony133:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

Issue Number: N/A

## What is the new behavior?
Add warning for use provider `APP_GUARD`.

- [ ] Yes
- [x] No

## Other information

Many people reading the documentation about guards find some confusion when they have to use a global guard using the provider (`APP_GUARD`), having the doubt that they also have to use the `useGlobalGuards()` method of the Nest application instance in the `main.ts` file. On Discord there are many discussion and help posts, I think a warning can help to solve this "confusion" in reading the documentation regarding guards.

### DIFF
--- a/content/guards.md
+++ b/content/guards.md
@@ -142,6 +142,8 @@ export class AppModule {}
 > where the guard (`RolesGuard` in the example above) is defined. Also, `useClass` is not the only way of dealing with
 > custom provider registration. Learn more [here](/fundamentals/custom-providers).
 
+> warning **Warning** When using the provider syntax with `APP_GUARD`, it is not necessary to use the `useGlobalGuards()` method of the Nest application instance in the `main.ts` file.
+
 #### Setting roles per handler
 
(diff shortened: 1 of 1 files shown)


content/guards.md
```diff
@@ -142,6 +142,8 @@ export class AppModule {}
 > where the guard (`RolesGuard` in the example above) is defined. Also, `useClass` is not the only way of dealing with
 > custom provider registration. Learn more [here](/fundamentals/custom-providers).
 
+> warning **Warning** When using the provider syntax with `APP_GUARD`, it is not necessary to use the `useGlobalGuards()` method of the Nest application instance in the `main.ts` file.
```
micalevisk, line 145: instead of adding this warning (that sounds more like an _info_), what about changing the above sentence

> In order to solve this issue, you can set up a guard directly from any module using the following construction:

to something that clearify indicates that we should replace `app.useGlobalGuards()` with the `APP_GUARD` provider? :thinking:

Tony133: It could also go, I preferred to add a warning so as not to touch the already documented part that seems very good to me.

kamilmysliwiec: I think this should be clear & obvious. No need to add an extra warning for this

PR #2600, opened 2023-01-29
docs(crud-generator.md): more descriptive shell usage

Nico2018:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

- [ ] Yes
- [x] No

### DIFF
--- a/content/recipes/crud-generator.md
+++ b/content/recipes/crud-generator.md
@@ -24,7 +24,7 @@ To help speed up this repetitive process, [Nest CLI](/cli/overview) provides a g
 To create a new resource, simply run the following command in the root directory of your project:
 
 ```shell
-$ nest g resource
+$ nest g resource <resource-name>
 ```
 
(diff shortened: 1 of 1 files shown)


content/recipes/crud-generator.md
```diff
@@ -24,7 +24,7 @@ To help speed up this repetitive process, [Nest CLI](/cli/overview) provides a g
 To create a new resource, simply run the following command in the root directory of your project:
 
 ```shell
-$ nest g resource
+$ nest g resource <resource-name>
```
micalevisk, line 27: I don't think this is needed as the resource name is not mandatory when invoking the command. Also, if one want to see how `resource` they should use `resource -h` or `resource --help`. So if we have that `<resource-name>`, we end up hidding other parameters:  
```
nest generate|g [options] <schematic> [name] [path]
```

Issue #2635, opened 2023-02-26
Application does not start on windows 10


Kee-Reall:
## I'm submitting a...

- [ ] Regression 
- [x] Bug report
- [ ] Feature request
- [ ] Documentation issue or request (new chapter/page)
- [ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.

## Current behavior
```
PS E:\Projects\docs-nest\src\app\homepage\pages> npm run start

> nestjs-docs@1.0.0 prestart
> npm run docs


> nestjs-docs@1.0.0 docs
> npm run docs-only


> nestjs-docs@1.0.0 docs-only
> ts-node -P tools/tsconfig.json tools/dgeni-cli.ts tools/transforms/nestjs-package/index

info:    running processor: readFilesProcessor
info:    running processor: extractJSDocCommentsProcessor
info:    running processor: parseTagsProcessor
info:    running processor: extractTagsProcessor
info:    running processor: codeNameProcessor
info:    running processor: computeIdsProcessor
info:    running processor: computeWhoUsesProcessor
info:    running processor: computePathsProcessor
info:    running processor: renderDocsProcessor
info:    running processor: unescapeCommentsProcessor
info:    running processor: inlineTagProcessor
info:    running processor: postProcessHtml
Illegal characters in path.
error:   Failed to process the docs
Error: Error running processor "cleanGeneratedFiles":
Illegal characters in path.
    at pathArg (E:\Projects\docs-nest\node_modules\rimraf\src\path-arg.ts:48:27)
    at AsyncFunction.<anonymous> (E:\Projects\docs-nest\node_modules\rimraf\src\index.ts:56:19)
    at CleanGeneratedFiles.$process (E:\Projects\docs-nest\tools\transforms\nestjs-package\processors\cleanGeneratedFiles.ts:12:12)
    at E:\Projects\docs-nest\node_modules\dgeni\src\Dgeni.ts:207:31 {
  path: 'E:\\Projects\\docs-nest\\src\\app\\homepage\\pages\\{docs,*.json}',
  code: 'EINVAL'
}
```
issue is [in](https://github.com/nestjs/docs.nestjs.com/blob/master/tools/transforms/nestjs-package/processors/cleanGeneratedFiles.ts) 
`\tools\transforms\nestjs-package\processors\CleanGeneratedFiles.ts`

```
// $process() {
[...]

PR #2668, opened 2023-03-19
fix(docs): asset cleaning errors on windows

tsa96:
## PR Checklist

## PR Type
[rimraf v4 recently removed support for glob patterns](https://github.com/isaacs/rimraf/issues/249), then subsequently added them back, but requiring the `{ glob: true }` in the options passed to rimraf API methods. My understanding is, on POSIX-compliant OSs, it'll use the native shell handling of glob patterns anyway, but on Windows, now needs to passed this option for it to use the [node-glob](https://github.com/isaacs/node-glob), which handles glob patterns on Windows.

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] Docs

As mentioned in the below issue, the `rimraf.sync` call in `cleanGeneratedFiles.ts` will error with "Illegal characters in path." when the `${OUTPUT_PATH}/{docs,*.json}` glob is passed in, apparently only on Windows (I'm on Windows 11 and get the exact same output as #2635).

Issue Number: #2635

## What is the new behavior?
Explicitly uses the built-in glob handling, which now runs fine on Windows. This may change behavior on other OSs, which at least since rimraf v4 were doing glob handling using native shell functionality, but I assume that getting identical behavior between different operating systems is why you're using rimraf in the first place.

- [ ] Yes
- [x] No

### DIFF
--- a/tools/transforms/nestjs-package/processors/cleanGeneratedFiles.ts
+++ b/tools/transforms/nestjs-package/processors/cleanGeneratedFiles.ts
@@ -9,7 +9,7 @@ export class CleanGeneratedFiles implements Processor {
   $runAfter = ['writing-files'];
   $runBefore = ['writeFilesProcessor'];
   $process() {
-    rimraf.sync(`${OUTPUT_PATH}/{docs,*.json}`);
+    rimraf.sync(`${OUTPUT_PATH}/{docs,*.json}`, { glob: true });
   }
 }
 


micalevisk (approved):
I tested `npm run build:watch` on Ubuntu.

thanks!

kamilmysliwiec:
lgtm

PR #2691, opened 2023-04-04
docs(caching): added warning for redis store

Tony133:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

Issue Number: N/A

## What is the new behavior?
Proposal:

I thought I would add this warning because the code snippet of the cache manager with Redis described in the documentation confuses people a bit and they come to think that it also works with the v5 cache manager, instead it only works with the v4 version. (see screenshot).

screenshot-config-redis-cache

Momentarily I think it is a good thing to add a warning. 

**So in the meantime we can prepare an 'update of the "Different Stores" and "Asynchronous Configuration" chapter of the "Caching" section because the TTL with the new configurations is set directly within the configuration (see code snippet) rather than as a module option**

```ts
CacheModule.registerAsync({
  useFactory: async () => {
      const store = await redisStore({
		socket: {
		  host: 'localhst',
		  port: '6379',
		  tls: true
		},
		password: 'redis-pass',
     });

     return {
	store: store as unknown as CacheStore,
	ttl: 60 * 60 * 24 * 7,  // 1 week
      };
   } 
})

```

- [ ] Yes
- [x] No

### DIFF
--- a/content/techniques/caching.md
+++ b/content/techniques/caching.md
@@ -246,6 +246,8 @@ import { AppController } from './app.controller';
 export class AppModule {}
 ```
 
+> warning**Warning** This snippets for configuration with Redis does not work with cache manager v5 but only with v4 version.
+
 > warning**Warning** `cache-manager-redis-store` does not support redis v4. In order for the `ClientOpts` interface to exist and work correctly you need to install the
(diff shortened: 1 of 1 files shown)


Tony133:
I close because it is replaced by this https://github.com/nestjs/docs.nestjs.com/pull/2703

PR #2724, opened 2023-05-06
docs(vitest): explicitly specify swcrc option to avoid conflict with vite

thenbe:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

Error when running `vitest` with an existing `.swcrc` as outlined in the `swc` recipe.

As it stands, both the `swc` and `vitest` recipes work great in their own right. When used together however, the module type needs to be explicitly set in `vitest.config.ts`. Otherwise it will be inherited from the `.swcrc`, which leads to an error `Cannot find module...`.

Issue Number: N/A

## What is the new behavior?

No error when running `vitest` with an existing `.swcrc` as outlined in the `swc` recipe.

- [ ] Yes
- [x] No

## Other information

Thanks for adding these useful recipes.

### DIFF
--- a/content/recipes/swc.md
+++ b/content/recipes/swc.md
@@ -121,7 +121,13 @@ export default defineConfig({
     globals: true,
     root: './',
   },
-  plugins: [swc.vite()], // This is required to build the test files with SWC
+  plugins: [
+    // This is required to build the test files with SWC
+    swc.vite({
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
lgtm

PR #2784, opened 2023-07-05
docs(caching): add note about of cache-manager

jpham005:
Since cache manager's In-memory store uses `lodash.clonedeep` internally, it can only store values of types supported by 'lodash'.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

- Property descriptors, setters, getters, and similar metadata-like features are not duplicated. For example, if an object is marked readonly with a [property descriptor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyDescriptor), it will be read/write in the duplicate, since that's the default. [mdn-Structured_clone_algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm)

```ts
import { ObjectId } from "bson"; // import package to test invalid type which uses 
import { CACHE_MANAGER } from '@nestjs/cache-manager';
import { Cache } from 'cache-manager';

export class ExampleService {
  constructor (
    @Inject(CACHE_MANAGER)
    private cacheManagerService: Cache;
  ) {}

  async generatesNodeInternalError() {
    const objectId = new ObjectId("some id..."); // has OwnPropertyDescriptors
    const map = new Map();
    map.set('mapKey', objectId);  
    
    const cacheKey = 'cacheKey';

    await this.cacheManagerService.set(cacheKey, map);
    // Since `console.log` uses custom inspect function of ObjectId,
    // which requires properties in original object, an exception will occur below.
    console.log(await this.cacheManagerService.get(cacheKey));
  }
}

```

Issue Number: N/A

- [ ] Yes
- [x] No

## Other information
1. `@nestjs/cache-module` is not passing `isCachable` to `node-cache-manager`. [link](https://github.com/nestjs/cache-manager/blob/master/lib/cache.providers.ts)
2. `node-cache-manager` clones value by `lodash.clonedeep`. [link](https://github.com/node-cache-manager/node-cache-manager/blob/master/src/stores/memory.ts#L6)
3. Then call `set` of lru-cache. After here, cached value's PropertyDescriptors are omitted. [link](https://github.com/node-cache-manager/node-cache-manager/blob/master/src/stores/memory.ts)

### DIFF
--- a/content/techniques/caching.md
+++ b/content/techniques/caching.md
@@ -17,6 +17,8 @@ $ npm install @nestjs/cache-manager cache-manager
 
 #### In-memory cache
 
+> warning **Note** `In-memory cache` can only store values of types that are supported by [the structured clone algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#javascript_types).
+
 Nest provides a unified API for various cache storage providers. The built-in one is an in-memory data store. However, you can easily switch to a more comprehensive solution, like Redis.
 
 In order to enable caching, import the `CacheModule` and call its `register()` method.


content/techniques/caching.md
```diff
@@ -17,6 +17,8 @@ $ npm install @nestjs/cache-manager cache-manager
 
 #### In-memory cache
 
+> warning **Note** `In-memory cache` can only store values available in [The structured clone algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm).
```
micalevisk, line 20:
```suggestion
> warning **Note** `In-memory cache` can only store values available that are supported by [the structured clone algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#javascript_types).
```

jpham005: 👍

micalevisk (approved):

PR #2797, opened 2023-07-12
fix: remove outdated section about shutdown hooks with prisma

aqrln:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

Current Prisma guide needs to be updated as custom `enableShutdownHooks` method has not been necessary since Prisma 3, and now the example doesn't compile since Prisma 5.

This leads to confusion for users: https://github.com/prisma/prisma/issues/20171
Release notes: https://github.com/prisma/prisma/releases/tag/5.0.0

## What is the new behavior?

The outdated section is removed.

- [ ] Yes
- [x] No

## Other information

This section has not been relevant since the introduction of the library engine as the default in Prisma 3. Now, in Prisma 5, the code example does not compile anymore since the `beforeExit` event was removed from Prisma Client (except when generating the client with the binary engine type, which most users should not do unless they have good reasons).

Note that even in Prisma 2 or when switching to the binary engine, using `beforeExit` would've only been necessary to be able to run database queries in lifecycle hooks, and wouldn't be necessary otherwise. The justification for using `beforeExit` that was present in this section was not fully accurate because while the bug it refers to (https://github.com/prisma/prisma/issues/3773) was fixed together with the introduction of `beforeExit` hook (which I believe led to the confusion), those were not strictly related changes.

### DIFF
--- a/content/recipes/prisma.md
+++ b/content/recipes/prisma.md
@@ -248,7 +248,7 @@ When setting up your NestJS application, you'll want to abstract away the Prisma
 Inside the `src` directory, create a new file called `prisma.service.ts` and add the following code to it:
 
 ```typescript
-import { INestApplication, Injectable, OnModuleInit } from '@nestjs/common';
+import { Injectable, OnModuleInit } from '@nestjs/common';
 import { PrismaClient } from '@prisma/client';
 
 @Injectable()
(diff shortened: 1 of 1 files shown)


content/recipes/prisma.md
```diff
@@ -248,24 +248,18 @@ When setting up your NestJS application, you'll want to abstract away the Prisma
...
-
-  async enableShutdownHooks(app: INestApplication) {
-    this.$on('beforeExit', async () => {
-      await app.close();
-    });
-  }
 }
 ```
 
-> info **Note** The `onModuleInit` is optional — if you leave it out, Prisma will connect lazily on its first call to the database. We don't bother with `onModuleDestroy`, since Prisma has its own shutdown hooks where it will destroy the connection. For more info on `enableShutdownHooks`, please see [Issues with `enableShutdownHooks`](recipes/prisma#issues-with-enableshutdownhooks)
```
aqrln, line 268: > We don't bother with `onModuleDestroy`, since Prisma has its own shutdown hooks where it will destroy the connection.

This was also a little inaccurate, at least since Prisma 3.

Prisma's own shutdown hooks in Prisma 2 (or in modern Prisma when using the binary engine) existed to kill the query engine process, but they didn't explicitly close the db connection. I'm not sure if the engine did that by catching the signal itself, but I can't seem to find that in the engines codebase.

In Prisma 3+, with library engine (a Node.js native addon), the only reason the internal shutdown hooks still existed was to maintain backward compatibility and keep the `beforeExit` event, but they haven't been doing anything beyond that.

If it's desired to close the connections gracefully rather than dropping them, it's still necessary to add an `onModuleDestroy` method with explicit `$disconnect`.

Jolg42 (approved):

micalevisk (approved):

kamilmysliwiec:
LGTM

PR #2837, opened 2023-08-28
docs: update auth guard logic to facilitate both public and private implementation on a single route

AIM3r4j:
This change opens the path to implement different logic based on the access level on that single route.


kamilmysliwiec:
https://github.com/nestjs/nest/pull/12287

PR #2859, opened 2023-09-12
Update swc.md: @vitest/coverage-c8 to @vitest/coverage-v8

leonardfreitas:
Just an update to the dependency name


jmcdo29 (approved):
Thanks for updating the name after the deprecation. LGTM

Tony133 (approved):

PR #2907, opened 2023-11-21
docs (logger): add missing Injectable decorator to custom logger service class

sebastiendan:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/content/techniques/logger.md
+++ b/content/techniques/logger.md
@@ -53,6 +53,7 @@ Implementing your own custom logger is straightforward. Simply implement each of
 ```typescript
 import { LoggerService } from '@nestjs/common';
 
+@Injectable()
 export class MyLogger implements LoggerService {
   /**
    * Write a 'log' level log.


jmcdo29 (commented):
_Technically_ this isn't needed. There's nothing that is injected into the `Logger` class, so the `@Injectable()` here is extraneous. We can still add it for continuity, but it isn't explicitly necessary

sebastiendan:
> _Technically_ this isn't needed. There's nothing that is injected into the `Logger` class, so the `@Injectable()` here is extraneous. We can still add it for continuity, but it isn't explicitly necessary

I see. In that case another proposed change could be adding a mention to the need for the decorator in the later `Dependency injection` section where `ConfigService` is referred as an example (but perhaps it is expected that the reader knows any dependency injected in the class constructor requires the `Injectable` class decorator?).

kamilmysliwiec:
https://github.com/nestjs/docs.nestjs.com/pull/2907#pullrequestreview-1742799827

kamilmysliwiec:
I figured let's just add it for simplicity sake

PR #2991, opened 2024-03-13
docs(exception-filters): correct status property

tomaszwysocki:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

In #2988 status has been changed to statusCode, but it's inconsistent with an example:

image

I think it should be changed back to 'status,' or 'status' in the aforementioned example should be changed to 'statusCode.' It doesn't matter which. The case that @wilsonyiyi mentioned is irrelevant here. The docs say that we can pass an object, and there is no mention that it has to own a property called 'statusCode'.
> To override just the message portion of the JSON response body, supply a string in the response argument. To override the entire JSON response body, pass an object in the response argument. Nest will serialize the object and return it as the JSON response body.

## What is the new behavior?
Object properties used in the example will now remain consistent with the documentation.

- [ ] Yes
- [x] No

### DIFF
--- a/content/exception-filters.md
+++ b/content/exception-filters.md
@@ -72,7 +72,7 @@ async findAll() {
     await this.service.findAll()
   } catch (error) { 
     throw new HttpException({
-      statusCode: HttpStatus.FORBIDDEN,
+      status: HttpStatus.FORBIDDEN,
       error: 'This is a custom message',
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
Thank you @tomaszwysocki

PR #3006, opened 2024-04-02
docs(pipes): replace the word "metatype" with "metadata"

gyt95:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

The word "metatype" should be replaced by "metadata" because the second parameter of the transform method is metadata, and the meaning here should be to deconstruct the metatype field and assign it to the second parameter metadata of transform.

## What is the new behavior?
By making this adjustment, the correct meaning can be expressed.

- [ ] Yes
- [x] No

### DIFF
--- a/content/pipes.md
+++ b/content/pipes.md
@@ -398,7 +398,7 @@ export class ValidationPipe implements PipeTransform<any> {
 
 Let's go through this code. First, note that the `transform()` method is marked as `async`. This is possible because Nest supports both synchronous and **asynchronous** pipes. We make this method `async` because some of the class-validator validations [can be async](https://github.com/typestack/class-validator#custom-validation-classes) (utilize Promises).
 
-Next note that we are using destructuring to extract the metatype field (extracting just this member from an `ArgumentMetadata`) into our `metatype` parameter. This is just shorthand for getting the full `ArgumentMetadata` and then having an additional statement to assign the metatype variable.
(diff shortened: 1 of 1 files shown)


content/pipes.md
```diff
@@ -398,7 +398,7 @@ export class ValidationPipe implements PipeTransform<any> {
 
 Let's go through this code. First, note that the `transform()` method is marked as `async`. This is possible because Nest supports both synchronous and **asynchronous** pipes. We make this method `async` because some of the class-validator validations [can be async](https://github.com/typestack/class-validator#custom-validation-classes) (utilize Promises).
 
-Next note that we are using destructuring to extract the metatype field (extracting just this member from an `ArgumentMetadata`) into our `metatype` parameter. This is just shorthand for getting the full `ArgumentMetadata` and then having an additional statement to assign the metatype variable.
+Next note that we are using destructuring to extract the metatype field (extracting just this member from an `ArgumentMetadata`) into our `metadata` parameter. This is just shorthand for getting the full `ArgumentMetadata` and then having an additional statement to assign the metatype variable.
```
jmcdo29, line 401: We are explaining how destructuring works here. You are correct that the second parameter _under normal circumstances_ would be the `metadata` parameter; however, as we're destructuring `metatype` from that parameter, we're explicitly calling it the `metatype` parameter in the description here

gyt95: That's it, okay I understand.

PR #3068, opened 2024-08-01
docs: convert absolute paths to relative paths

### LINKED ISSUE #3067: Convert absolute paths to relative paths in documentation
## I'm submitting a...

- [ ] Regression 
- [ ] Bug report
- [ ] Feature request
- [x] Documentation issue or request (new chapter/page)
- [ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.

## Current behavior

Within the document, there are internal links using relative paths, while there are also internal links using absolute paths.

## Expected behavior

Internal links within the document should use relative paths.

I believe that the absolute paths should be changed to relative paths for the following two reasons:
1. Consistency
2. When translating and distributing the document, there is an issue where the links in the translated document point back to the original English document. To resolve this, the translator would need to manually convert the links, but a better solution is to use relative paths.

bombard1004:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

Issue Number: closes #3067 

- [ ] Yes
- [x] No


kamilmysliwiec:
We intentionally use absolute paths for links that should open in a new browser tab (with target = blank)

bombard1004:
Hi, thank you for sharing the context.

After further investigation, I’ve found that the common practice is to append {:target="_blank"} to Markdown links to open them in a new tab. Unfortunately, it appears that the `markedjs` library does not support parsing this syntax.

I also noticed that the document mixes Markdown links and HTML anchor tags. Given this, I’d like to suggest using HTML anchor tags for internal links that need to open in a new tab, allowing the use of target="_blank". I would appreciate your thoughts on this approach.

kamilmysliwiec:
Are we sure this is worth the effort though? We don't plan to add any translations to the docs any time soon (and possibly ever).

PR #3080, opened 2024-08-24
docs: Fix cache-manager TTL documentation

mohamedelraghy:
### Issue Description
The NestJS Cache Manager documentation states that TTL should be passed in milliseconds when using `await this.cacheManager.set('key', 'value', 1000);`. However, in practice, this does not work. Instead, passing the TTL as part of an options object in seconds (`{ ttl: <seconds> }`) works correctly.

### Steps to Reproduce:
1. Follow the current documentation to pass the TTL in milliseconds. The cache key is not set as expected.
2. Using `{ ttl: 1 }` in seconds successfully sets the cache key.

### Fix:
Updated the documentation to reflect that the TTL should be passed in seconds via an options object instead of milliseconds.

### DIFF
--- a/content/techniques/caching.md
+++ b/content/techniques/caching.md
@@ -62,7 +62,7 @@ The default expiration time of the cache is 5 seconds.
 You can manually specify a TTL (expiration time in seconds) for this specific key, as follows:
 
 ```typescript
-await this.cacheManager.set('key', 'value', 1000);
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
Depends on what version you use. For `cache-manager>=5` it should be `this.cacheManager.set('key', 'value', 1000)`

PR #3097, opened 2024-09-25
docs(documentation): updated swc documentation with gotcha

GianlucaCesari:
Updated swd.md to add the example on how to import the type Relation from typeorm

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

Issue Number: N/A

- [ ] Yes
- [x] No

## Other information

Added a gotcha to the swc documentation.

The Relation type should be imported as a type instead of directly from typeorm as happens when you use the auto import in the ides. I think it should be documented so nobody loses 45min debugging this again (💀)

### DIFF
--- a/content/recipes/swc.md
+++ b/content/recipes/swc.md
@@ -178,7 +178,10 @@ export class User {
 }
 ```
 
-> info **Hint** `Relation` type is exported from the `typeorm` package.
+> info **Hint** `Relation` type is exported from the `typeorm` package and should be imported as a type.
+> ```typescript
+>  import type { Relation } from 'typeorm';
+> ```
 
 Doing this prevents the type of the property from being saved in the transpiled code in the property metadata, preventing circular dependency issues.
 


kamilmysliwiec:
Even if you don't use `import type` you should be fine

micalevisk:
I saw in typeorm's codebase that `Relation` is a type already (here https://github.com/typeorm/typeorm/blob/master/src/common/RelationType.ts) so `import type` is redundant.

GianlucaCesari:
Ok, just for reference, when i omit the import type this error pops up, maybe it's a configuration error on my part, just writing it down if someone else stumbles upon this error

```bash
import { Column, CreateDateColumn, Entity, Index, PrimaryGeneratedColumn, Relation, UpdateDateColumn } from "typeorm";
                                                                          ^^^^^^^^
SyntaxError: The requested module 'typeorm' does not provide an export named 'Relation'

```

And i need to import it as

```typescript
import {
  Column,
  CreateDateColumn,
  Entity,
  Index,
  PrimaryGeneratedColumn,
  type Relation,
  UpdateDateColumn,
} from 'typeorm';
```

To make it work.

For reference this is my .swrc config


```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "decorators": true,
      "dynamicImport": true
    },
    "target": "es2019",
    "transform": {
      "decoratorMetadata": true
    }
  },
  "module": {
    "type": "es6",
    "strict": true,
    "strictMode": true,
    "noInterop": false
  },
  "sourceMaps": true
}

```

PR #3102, opened 2024-10-09
feat: respect user's preferred color scheme (dark mode repair)

jewhyena:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] Docs

Currently, the Nest.js documentation site does not respect the user's system preferences regarding the color scheme (light or dark). As a result, opening the site in a new browser or in an Incognito session always defaults to light mode, ignoring the system preferences.

Manually toggling the theme will save the current theme in `localStorage`, so the next time the user opens the site, it will remain the same.

Issue Number: N/A

## What is the new behavior?
The new behavior respects the user's system preferences. If their preferred color scheme is dark, the Nest.js site will default to dark mode, and similarly, if their preferred color scheme is light, it will be in light mode. It now also listens for changes in the preferred color scheme, and the Nest.js site will automatically change its theme according to the current system preferences.

If the user manually toggles the theme button, the theme will be stored in `localStorage` and remain there. In this case, the theme will not respect the system preferences. Instead, `localStorage` will be the source of truth.

Please see the attached video.

- [ ] Yes
- [x] No

## Other information

https://github.com/user-attachments/assets/3656f7de-3dfc-419f-9ced-51853460f9f5

### DIFF
--- a/src/app/shared/components/theme-mode-toggle/theme-mode-toggle.component.ts
+++ b/src/app/shared/components/theme-mode-toggle/theme-mode-toggle.component.ts
@@ -21,18 +21,18 @@ export class ThemeModeToggleComponent implements OnInit {
   ) {}
 
   ngOnInit() {
-    const preferredScheme = this.mediaMatcher.matchMedia(
+    const darkSchemeMatcher = this.mediaMatcher.matchMedia(
       '(prefers-color-scheme: dark)',
     );
 
-    preferredScheme.onchange = () => {
-      if (!this.getStoredTheme()) this.toggleTheme(true);
+    darkSchemeMatcher.onchange = ({ matches }) => {
+      if (!this.getStoredTheme()) this.setTheme(matches ? 'dark' : 'light');
     };
 
-    const isDarkSchemePreferred = preferredScheme.matches;
+    const preferredScheme = darkSchemeMatcher.matches ? 'dark' : 'light';
     const storedTheme = this.getStoredTheme();
 
-    this.theme = storedTheme ?? (isDarkSchemePreferred ? 'dark' : 'light');
+    this.theme = storedTheme ?? preferredScheme;
     this.setTheme(this.theme);
   }
 
--- a/src/app/shared/components/theme-mode-toggle/theme-mode-toggle.component.ts
+++ b/src/app/shared/components/theme-mode-toggle/theme-mode-toggle.component.ts
@@ -3,13 +3,15 @@ import { DOCUMENT } from '@angular/common';
 import { MediaMatcher } from '@angular/cdk/layout';
 import { StorageService } from '../../services/storage.service';
 
+type Theme = 'light' | 'dark';
+
 @Component({
   selector: 'app-theme-mode-toggle',
   templateUrl: './theme-mode-toggle.component.html',
   styleUrls: ['./theme-mode-toggle.component.scss'],
 })
 export class ThemeModeToggleComponent implements OnInit {
-  isDarkMode: boolean;
+  theme: Theme;
 
   constructor(
     @Inject(DOCUMENT)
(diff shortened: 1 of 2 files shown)


micalevisk:
I guess this isn't working on Ubuntu (firefox) :thinking: 

[demo_linux.webm](https://github.com/user-attachments/assets/3cc49f5c-ac41-4a8a-abe1-1afc2e95f0f7)

jewhyena:
> I guess this isn't working on Ubuntu (firefox) 🤔 I tested the deploy preview
> 
>  demo_linux.webm

Hm. That is strange. There's no platform-dependent code.

It works seamlessly on both Chrome and Firefox on my Mac, as well as on my iPhone.

By any chance, didn't you manually change the theme on the site before starting switching your system preference? Was your `localStorage` clear?

Because once you manually change the theme on the site, it no longer respects your system preference. After you click the button, the `localStorage` becomes the source of truth for the theme. This behavior is expected.

micalevisk:
still not working here :thinking: 

[demo2.webm](https://github.com/user-attachments/assets/287d61f9-af87-4c62-8766-f4d15b275090)

while the theme switching worked for discord desktop app

jewhyena:
> while the theme switching worked for discord desktop app

Hmm.

Since Discord is a different app which might not have any issues resolving the preferred schema on Ubuntu, let's check if it's the same with https://reactnative.dev/ and https://reactnavigation.org/ websites.

Could you please tell me whether those two sites respect your color scheme or do nothing as well?

(`localStorage` must be clean on those sites as well.)

micalevisk:
they didn't work on firefox neither

but I just found that everything went fine on chrome

jewhyena:
> they didn't work on firefox neither
> 
> but I just found that everything went fine on chrome

Might be some issue with GNOME: https://askubuntu.com/a/1464951. 🤔

src/app/shared/components/theme-mode-toggle/theme-mode-toggle.component.ts
```diff
@@ -1,49 +1,57 @@
...
     private readonly storageService: StorageService,
+    private readonly changeDetector: ChangeDetectorRef,
   ) {}
 
   ngOnInit() {
-    // This is commented out because by default the theme mode is set to light (at least for now)
-    // const userPrefersTheme =
-    //   this.mediaMatcher.matchMedia &&
-    //   this.mediaMatcher.matchMedia('(prefers-color-scheme: light)').matches;
-    // this.setThemeMode(this.getUserSettingsIsDarkMode() || userPrefersTheme);
```
kamilmysliwiec, line 26: Wouldn't just uncommenting this code be enough?

jewhyena: This is what I originally hoped for, but it didn't seem to be working. So I started fixing it.

https://github.com/user-attachments/assets/45827489-7940-4393-99ef-d7c199e6b5f7

kamilmysliwiec: Are you certain there's nothing in your local storage that specifies which mode to use? If I remember correctly, the commented-out code checks local storage first and, if it's empty, falls back to the user's preferred color scheme.

jewhyena: > Are you certain there's nothing in your local storage that specifies which mode to use?

Yes, absolutely. You can see on the video that my `localStorage` is empty.

Moreover, looking at the old code, it isn't even supposed to listen to any changes in color scheme preferences, since there is no `onChange` event listener specified anywhere. It only supposed to get the system preferences on the initial load (`ngOnInit`), but it turns out that it also doesn't work, if we just uncomment the code. I get the light mode despite of my system preference being dark.

It might seem like the code on lines 28-29, which overrides everything returned by the `mediaMatches`, was put in there as a temporary substitution for the commented code. If we remove that and leave only uncommented code, it starts working but in the opposite direction: if my system preference is dark, it gives me the light mode and, vice versa, if my system preference is light, only then it gives me the dark mode.

However, if we change the `mediaMatcher` query from `(prefers-color-scheme: light)` to `(prefers-color-scheme: dark)`, it starts to give me the right theme.

But it still doesn't listen to any changes in the system preferences. If you change your system preference manually or if your OS configured to change the system preference automatically at different times of the day, you **have to** refresh your page in order for theme to change. This is what I intended to implement in this PR.

I added an event listener, which, if there's no stored theme in `localStorage`, would automatically change the theme according to the system preference, with no need to refresh the page. Also there's `skipStorage` parameter which allows skipping saving the theme in `localStorage`. We need this flag in order to prevent `onChange` event from storing anything in `localStorage`. Otherwise, once our system preference changes for the first time, the site will no longer respect it, despite the user not selecting the preferred theme manually by clicking the theme button. This is not the behavior we usually expect and I'd consider it a bug. So this is already handled.

What I'm not so sure about, though, is the use of `changeDetectorRef`. I don't know if this is a good or a bad practice, since this PR is basically the first time I ever touch Angular. So I would appreciate any advice regarding it.

What I've added it for is that when the system preference change, the theme icon in the top right corner stays the same. You can notice it on the video I first attached to the PR. I recorded it before noticing it myself and before I pushed the fix.

For some reason, when `theme` change was caused by `darkSchemeMatcher.onchange` event, `*ngIf`s in the `.html` file didn't react to the change and stayed the same. So I found this approach with `changeDetectorRef` and with it all went as expected.

kamilmysliwiec:
LGTM

PR #3193, opened 2025-02-03
docs(resolvers-map): add notes about missing non-value types for nullable fields and args

smolinari:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [X] Docs

## What is changing?

There is often confusion as to why adding `| null` or even `| null | undefined` to nullable fields or args in resolver definitions causes an error. This change attempts to help make this "way" of keeping the types with GraphQL nullable fields or args distinct a bit more clearer and to point to what is expected by the developer, when working with nullable fields.  

- [ ] Yes
- [X] No

Scott

### DIFF
--- a/content/graphql/resolvers-map.md
+++ b/content/graphql/resolvers-map.md
@@ -269,7 +269,7 @@ getAuthor(
 ) {}
 ```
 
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
lgtm

PR #3258, opened 2025-05-09
docs(migration): add slash for Express named wildcard

sjoerdsmink:
Add slash for Express route matching syntax

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

Incorrect mentioning of `{*splat}` that it is a named wildcard that matches any path including the root path. It actually does not include the root path. See the [Express migration docs](https://expressjs.com/en/guide/migrating-5.html#path-syntax)

Issue Number: N/A

## What is the new behavior?
Adding a slash in the example, and mentioning that it doesn't include the root path.

- [ ] Yes
- [x] No

### DIFF
--- a/content/migration.md
+++ b/content/migration.md
@@ -51,10 +51,10 @@ forRoutes('*'); // <-- This should not work in Express v5
 Instead, you can update the path to use a named wildcard:
 
 ```typescript
-forRoutes('{*splat}'); // <-- This will work in Express v5
+forRoutes('/{*splat}'); // <-- This will work in Express v5
 ```
 
-Note that `{{ '{' }}*splat&#125;` is a named wildcard that matches any path including the root path. Outer braces make path optional.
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
`/` shouldn't be required IIRC

sjoerdsmink:
@kamilmysliwiec It's mentioned explicitly in the [docs](https://expressjs.com/en/guide/migrating-5.html#path-syntax), in the note

> `*splat` matches any path without the root path. If you need to match the root path as well `/`, you can use `/{*splat}`, wrapping the wildcard in braces.

kamilmysliwiec:
Nest adds `/` automatically though

sjoerdsmink:
Ok, I didn't know that, thanks. `forRoutes` seems to work without the slash as well.

PR #3266, opened 2025-05-20
docs(versioning): remove redundant bracket

Uaitt:
## PR Checklist

## Description
There is a redundant `>`  bracket in the Versioning article. It's a result of commit https://github.com/nestjs/docs.nestjs.com/commit/a722d1e4b305be0b8a5bb04ba4065253ead37df6 and I think that it should be removed :)

screenshot

### DIFF
--- a/content/techniques/versioning.md
+++ b/content/techniques/versioning.md
@@ -107,7 +107,7 @@ For example, if an incoming request specifies it supports versions `1`, `2`, and
 If versions `[3, 2, 1]` are extracted, but routes only exist for version `2` and `1`, the route that matches version `2`
 is selected (version `3` is automatically ignored).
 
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
without > there's no space between these two 
image

Uaitt:
@kamilmysliwiec good catch 🤔.

It seems like this problem is already present in a few occurrences throughout other articles:
```sh
grep -r '` \*' --include="*.md" ./content
```

screenshot2

And in fact if we look at the actual webpage we see that the problem is there:

screenshot3

kamilmysliwiec:
If you'd like to submit a PR that resolves this issue at the compiler-level, that would be great!

PR #3281, opened 2025-06-15
docs: clarify lifecycle events

TheCoreMan:
I stumbled upon this in https://www.mrnice.dev/posts/nestjs-lifecycle/ and it confused me and my teammates. Thought it would be useful to clarify this

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

Issue Number: N/A

- [ ] Yes
- [x] No

## Other information

This is my first contrib to NestJS - tried to follow the guidelines, hoping I didn't miss anything. This was just a docs change so just did everything as quickly as possible from the web interface on GitHub


content/fundamentals/lifecycle-events.md
```diff
@@ -18,7 +18,7 @@ In the following table, `onModuleDestroy`, `beforeApplicationShutdown` and `onAp
 
 | Lifecycle hook method           | Lifecycle event triggering the hook method call                                                                                                                                                                   |
 | ------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `onModuleInit()`                | Called once the host module's dependencies have been resolved.                                                                                                                                                    |
+| `onModuleInit()`                | Called once the host module's dependencies have been resolved. "Once" meaning called AFTER the host module's dependencies have been resolved, but not necessarily one time only.                              |
```
kamilmysliwiec, line 21: Not sure if I'm following. This hook is called one time only _per each provider_ that implements it. The change you suggested ("not necessarily one time only") may lead someone to think that it might be called several times for the same provider (which isn't true). Correct me if I'm wrong

magoogli: @kamilmysliwiec I have had the same issue - I have a situation where the hook is called twice for the same provider. This scenario is easily reproduced. See https://github.com/magoogli/nest-js for a minimal repo. Is this expected behaviour? My onModuleInit hook is called twice for the same provider ( although the constructor is only called once ). This is very confusing and frustrating. Any help would be greatly appreciated.

PR #3427, opened 2026-05-08
docs(logger): explain why transient scope is used for custom logger

### LINKED ISSUE #3416: Missing explanation for `transient` in Logger DI docs
I think it would be worthwhile to explain the reason behind why would one want to have a unique instance of the logger in each feature module.

> We supply the `scope` option as configuration metadata for the `ConsoleLogger` class, specifying a [transient](/fundamentals/injection-scopes) scope, to ensure that we'll have a unique instance of the `MyLogger` in each feature module.

https://github.com/nestjs/docs.nestjs.com/blob/f44cfa1025a5b22071a35215f88917b61ca49f47/content/techniques/logger.md?plain=1#L292

pandaPowder:
## Problem

The "Injecting a custom logger" section states that `Scope.TRANSIENT` ensures a unique `MyLogger` instance per feature module, but doesn't explain *why* that uniqueness matters — leaving readers to connect the dots themselves.

Reported in #3416.

## Solution

Add two sentences explaining the motivation: each module calls `setContext()` on its logger to tag its log output with the module's name, and a singleton logger would cause those calls to overwrite each other across the app.

The existing code comment in the example already hints at this, so the new prose sets it up cleanly without being redundant.

Closes #3416

### DIFF
--- a/content/techniques/logger.md
+++ b/content/techniques/logger.md
@@ -289,7 +289,7 @@ You can also inject this `MyLogger` provider in your feature classes, thus ensur
 
 #### Injecting a custom logger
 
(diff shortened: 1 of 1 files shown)


micalevisk (approved):

kamilmysliwiec:
lgtm

PR #3435, opened 2026-05-20
docs(openapi): document OpenAPI 3.2 hierarchical tags

### LINKED ISSUE #3926: Documents using OpenAPI 3.2-only tag fields still declare openapi: 3.0.0
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

`SwaggerModule.createDocument()` builds the base document with a hard-coded `openapi: '3.0.0'`, overridable only when the user explicitly calls `DocumentBuilder.setOpenAPIVersion(...)`:

```ts
// lib/swagger-module.ts
const baseDocument: OpenAPIObject = {
  openapi: '3.0.0',
  paths: {},
  ...configWithoutWebhooks,
  ...documentWithoutWebhooks
};
```

Since v11.4.0 (#3725) the builder accepts OpenAPI **3.2-only** Tag Object fields — `parent` and `kind` (and `summary`, proposed in #3921):

```ts
new DocumentBuilder()
  .addTag('Cats', 'Cat operations', undefined, { parent: 'Animals', kind: 'nav' })
  .build();
```

The resulting document contains 3.2 fields but still declares:

```json
{ "openapi": "3.0.0", "tags": [{ "name": "Cats", "parent": "Animals", "kind": "nav" }] }
```

### Expected behavior

A document that uses 3.2-only constructs should not advertise itself as `3.0.0`. `parent`/`kind` are not valid Tag Object fields in 3.0.x, so strict validators (and 3.0-only tooling) flag the document as invalid — or silently drop the fields.

### Possible directions

Raising this as a design question rather than prescribing a fix:

1. **Auto-bump** — when any tag carries `parent`/`kind`/`summary`, set `openapi` to `3.2.0` unless the user pinned a version explicitly. Most convenient, but mutates the declared version implicitly.
2. **Warn** — keep the `3.0.0` default but emit a warning when 3.2-only fields are present and the version is `< 3.2.0`, pointing to `setOpenAPIVersion('3.2.0')`. Consistent with the approach chosen in #3922.
3. **Document only** — leave behavior as-is and note in the docs that `setOpenAPIVersion('3.2.0')` is required when using hierarchical tags.

Option 2 mirrors the resolution of #3922 and is the least surprising. Happy to open a PR once a direction is agreed.

### Package version

11.4.2

### Related

- #3725 — added `parent`/`kind` to the Tag Object
- #3921 — proposes adding `summary`
- #3922 — `@ApiTags` hierarchy fields

frbuceta:
## Summary

Adds a **Hierarchical tags** subsection to the *Operations → Tags* page, documenting the `parent` and `kind` options on `DocumentBuilder.addTag()` introduced for OpenAPI 3.2 support.

It also notes that `setOpenAPIVersion('3.2.0')` must be called when using these fields — otherwise the generated document still declares `openapi: 3.0.0` and strict validators reject the new tag fields.

## Context

Resolves nestjs/swagger#3926, where the maintainer chose the "document only" path: leave behavior as-is and document that `setOpenAPIVersion('3.2.0')` is required for hierarchical tags.

The subsection also mentions that hierarchy fields have no effect when passed to `@ApiTags()` — consistent with the runtime warning added in nestjs/swagger#3924.

### DIFF
--- a/content/openapi/operations.md
+++ b/content/openapi/operations.md
@@ -12,7 +12,6 @@ To attach a controller to a specific tag, use the `@ApiTags(...tags)` decorator.
 export class CatsController {}
 ```
 
-##### Hierarchical tags
 
 OpenAPI 3.2 extends the Tag Object so that tags can be organized into a hierarchy and annotated with a hint about how they should be presented. To declare these relationships, define the tags up front with `DocumentBuilder` and pass the `parent` and `kind` options to `addTag()`:
 
--- a/content/openapi/operations.md
+++ b/content/openapi/operations.md
@@ -12,6 +12,23 @@ To attach a controller to a specific tag, use the `@ApiTags(...tags)` decorator.
 export class CatsController {}
 ```
 
+##### Hierarchical tags
+
+OpenAPI 3.2 extends the Tag Object so that tags can be organized into a hierarchy and annotated with a hint about how they should be presented. To declare these relationships, define the tags up front with `DocumentBuilder` and pass the `parent` and `kind` options to `addTag()`:
+
+```typescript
+const config = new DocumentBuilder()
+  .setOpenAPIVersion('3.2.0')
+  .addTag('Animals', 'Everything about animals', undefined, { kind: 'nav' })
+  .addTag('Cats', 'Cat operations', undefined, { parent: 'Animals' })
+  .addTag('Dogs', 'Dog operations', undefined, { parent: 'Animals' })
+  .build();
+```
+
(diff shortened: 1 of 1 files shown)


content/openapi/operations.md
```diff
@@ -12,6 +12,23 @@ To attach a controller to a specific tag, use the `@ApiTags(...tags)` decorator.
 export class CatsController {}
 ```
 
+##### Hierarchical tags
```
kamilmysliwiec:
```suggestion
```

PR #3442, opened 2026-06-04
docs: correct pipe name FileParsePipe to ParseFilePipe in file upload guide

emmayusufu:
The file upload techniques page refers to `FileParsePipe` once, in the sentence introducing the validators snippet:

> To understand how these can be used in conjunction with the aforementioned `FileParsePipe`...

The class is actually `ParseFilePipe`. Every other mention in the same guide uses the correct name, and the code block right after this sentence instantiates `new ParseFilePipe({...})`. A reader searching for `FileParsePipe` in the API would come up empty, so this swaps the one stray reference to the real class name.


kamilmysliwiec:
lgtm

PR #3443, opened 2026-06-05
docs: fix typos and incorrect identifiers in code examples

emmayusufu:
While reading through the docs I found a handful of code examples and references that name things that do not match the surrounding code or the framework API. Each one would leave a reader copying the snippet with something that does not compile or that points at a class/method that does not exist. All are one-token fixes:

- **migration.md**: the comma-delimited `methods` string spells the verb `PATH` instead of `PATCH` (the array on the same line uses `PATCH`).
- **techniques/mongo.md**: three schema-factory snippets use `CatsSchema`, which is never defined; the schema is `CatSchema`. Prose also calls the discriminator class `SignedUpEvent`, but it is defined as `SignUpEvent`.
- **recipes/mongodb.md**: prose calls the schema `CatsSchema`; the file defines and uses `CatSchema`.
- **faq/request-lifecycle.md**: `app.useGlobalGuard()` should be `app.useGlobalGuards()`.
- **security/rate-limiting.md**: the `skipIf` rows reference `@SkipThrottler()`; the decorator is `@SkipThrottle()`.
- **microservices/custom-transport.md**: the import is `ClientTcp` while the class on the next line extends `ClientTCP`.
- **graphql/federation.md** and **graphql/schema-generator.md**: `PostsResolvers` and `UsersResolvers` are imported and listed as providers, but the classes are `PostsResolver` and `UsersResolver` (singular), consistent with `RecipesResolver` and `AuthorsResolver`.
- **recipes/nest-commander.md**: an import is missing the `from` keyword.
- **devtools/overview.md**: `instrospected` -> `introspected`.

Happy to split this into separate PRs if you'd prefer.

### DIFF
--- a/content/devtools/overview.md
+++ b/content/devtools/overview.md
@@ -40,7 +40,7 @@ export class AppModule {}
 
 > warning **Warning** The reason we are checking the `NODE_ENV` environment variable here is that you should never use this module in production!
 
-Once the `DevtoolsModule` is imported and your application is up and running (`npm run start:dev`), you should be able to navigate to [Devtools](https://devtools.nestjs.com) URL and see the instrospected graph.
+Once the `DevtoolsModule` is imported and your application is up and running (`npm run start:dev`), you should be able to navigate to [Devtools](https://devtools.nestjs.com) URL and see the introspected graph.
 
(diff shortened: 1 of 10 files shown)


kamilmysliwiec:
lgtm

PR #3464, opened 2026-06-29
docs: fix "class-base" -> "class-based" comment

EduardF1:
The JS variant comment was missing the trailing "d"; matches the TS variant above.


kamilmysliwiec:
lgtm

PR #3465, opened 2026-06-29
docs: update fastify.io links to fastify.dev

EduardF1:
Two Fastify links still use the retired www.fastify.io domain; the same file already uses fastify.dev elsewhere.


kamilmysliwiec:
lgtm

PR #3469, opened 2026-07-11
docs: explain abstract classes as di tokens

### LINKED ISSUE #3446: docs: add dedicated section explaining abstract class as DI token vs interface + manual tokens
## Documentation Request

Hi NestJS team! First, thank you for the amazing framework — we really appreciate the work you put into it.

## The Challenge

When onboarding team members or doing code reviews, I've found it genuinely hard to point to a single place in the docs that explains **why** NestJS uses `abstract class` as a DI token instead of a plain TypeScript `interface`.

The current docs cover custom providers and injection tokens, but the rationale behind choosing `abstract class` over `interface + string/Symbol token` is scattered across different sections. This makes it difficult to justify the recommendation to the team with a direct documentation link.

## Suggestion

A short, dedicated section — perhaps in the **Custom Providers** or **Fundamentals** chapter — covering:

1. Why `interface` cannot be used as a DI token (TypeScript erases interfaces at runtime)
2. The three alternatives: `string` token, `Symbol` token, and `abstract class` as token
3. When to prefer `abstract class` (no `@Inject()` decorator needed, single artifact, full type-check support)
4. When a plain `interface` is still the right choice (no DI involved)

Something like:

\`\`\`ts
// ❌ interface — erased at runtime, cannot be used as token
export interface LoggerService { log(msg: string): void }

// ✅ abstract class — survives compilation, works as token automatically
export abstract class LoggerService { abstract log(msg: string): void }

@Module({
  providers: [{ provide: LoggerService, useClass: PinoLoggerService }],
})
export class AppModule {}

// constructor — no @Inject needed
constructor(private logger: LoggerService) {}
\`\`\`

This small addition would make it much easier for teams to make consistent, well-informed decisions without having to piece together the reasoning from multiple pages.

Thanks for considering it!

Krishan27:
## PR Checklist
- [x] The commit message follows our guidelines
- [x] Docs have been added / updated

## PR Type

- [x] Documentation content changes

The custom providers documentation explains non-class provider tokens but does not explicitly cover why TypeScript interfaces cannot be used as DI tokens or when abstract classes can be used instead.

Closes #3446

## What is the new behavior?

Adds a focused section explaining:
- interfaces are erased at runtime and cannot be DI tokens by themselves
- string and symbol tokens can be used with `@Inject()`
- abstract classes survive compilation and can act as both contract and runtime token
- abstract class tokens support constructor injection without `@Inject()`
- when to prefer string/Symbol tokens, abstract classes, or plain interfaces

- [ ] Yes
- [x] No

## Validation

- `npm run docs` passed
- `npm run lint` failed on existing Angular lint errors in `src/app`, unrelated to this markdown-only change

### DIFF
--- a/content/fundamentals/dependency-injection.md
+++ b/content/fundamentals/dependency-injection.md
@@ -208,6 +208,85 @@ export class CatsRepository {
 
 While we directly use the string `'CONNECTION'` in the above examples for illustration purposes, for clean code organization, it's best practice to define tokens in a separate file, such as `constants.ts`. Treat them much as you would symbols or enums that are defined in their own file and imported where needed.
 
+#### Interfaces and abstract classes
+
+TypeScript types/interfaces are erased during compilation, so Nest can't reference them at runtime. This means an interface can describe the shape of a dependency, but it can't be used as a DI token by itself.
+
+Since Nest resolves providers by runtime tokens, use a string or `Symbol` token when registering a provider for an interface:
+
+```typescript
+export interface LoggerService {
+  log(message: string): void;
+}
+
+export const LOGGER_SERVICE = Symbol('LOGGER_SERVICE');
+
+@Injectable()
+export class PinoLoggerService implements LoggerService {
+  log(message: string) {
+    // implementation details
+  }
+}
+
+@Module({
+  providers: [
+    {
+      provide: LOGGER_SERVICE,
+      useClass: PinoLoggerService,
+    },
+  ],
+})
+export class AppModule {}
+```
+
+To inject this provider, pass that token to the `@Inject()` decorator:
+
+```typescript
+@Injectable()
+export class CatsService {
+  constructor(
+    @Inject(LOGGER_SERVICE)
+    private readonly logger: LoggerService,
+  ) {}
+}
+```
+
+Abstract classes, unlike interfaces, exist at runtime. You can use an abstract class as both the TypeScript contract and the DI token:
+
+```typescript
+export abstract class LoggerService {
+  abstract log(message: string): void;
+}
+
+@Injectable()
+export class PinoLoggerService implements LoggerService {
+  log(message: string) {
+    // implementation details
+  }
+}
+
+@Module({
+  providers: [
+    {
+      provide: LoggerService,
+      useClass: PinoLoggerService,
+    },
+  ],
+})
+export class AppModule {}
+```
+
+With an abstract class token, constructor-based injection can use the abstract class type directly and doesn't require `@Inject()`:
+
+```typescript
+@Injectable()
+export class CatsService {
+  constructor(private readonly logger: LoggerService) {}
+}
+```
+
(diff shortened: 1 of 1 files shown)


content/fundamentals/dependency-injection.md
```diff
@@ -208,6 +208,91 @@ export class CatsRepository {
...
 
+#### Interfaces and abstract classes
+
+TypeScript interfaces are removed during compilation, so Nest can't reference them at runtime. This means an interface can describe the shape of a dependency, but it can't be used as a DI token by itself.
+
+```typescript
+export interface LoggerService {
+  log(message: string): void;
+}
+```
```
micalevisk:
```suggestion
```

just because we have this same code bellow

content/fundamentals/dependency-injection.md
```diff
@@ -208,6 +208,91 @@ export class CatsRepository {
 
 While we directly use the string `'CONNECTION'` in the above examples for illustration purposes, for clean code organization, it's best practice to define tokens in a separate file, such as `constants.ts`. Treat them much as you would symbols or enums that are defined in their own file and imported where needed.
 
+#### Interfaces and abstract classes
+
+TypeScript interfaces are removed during compilation, so Nest can't reference them at runtime. This means an interface can describe the shape of a dependency, but it can't be used as a DI token by itself.
```
micalevisk:
```suggestion
TypeScript types/interfaces are erased during compilation, so Nest can't reference them at runtime. This means an interface can describe the shape of a dependency, but it can't be used as a DI token by itself.
```

content/fundamentals/dependency-injection.md
```diff
@@ -208,6 +208,91 @@ export class CatsRepository {
...
+    {
+      provide: LOGGER_SERVICE,
+      useClass: PinoLoggerService,
+    },
+  ],
+})
+export class AppModule {}
+```
+
+To inject this provider, pass the token to the `@Inject()` decorator:
```
micalevisk:
```suggestion
To inject this provider, pass that token to the `@Inject()` decorator:
```

content/fundamentals/dependency-injection.md
```diff
@@ -208,6 +208,91 @@ export class CatsRepository {
...
+With an abstract class token, constructor-based injection can use the abstract class type directly and doesn't require `@Inject()`:
+
+```typescript
+@Injectable()
+export class CatsService {
+  constructor(private readonly logger: LoggerService) {}
+}
+```
+
+Use string or symbol tokens when the DI token should be decoupled from a class artifact, when several providers share the same interface, or when exposing an abstract class as part of your API isn't desirable. Use abstract classes when one artifact should act as both the contract and the runtime token, and you prefer simpler constructor injection. A plain interface is still a good choice when the type is only used for compile-time checking and no DI token is needed.
```
micalevisk: I guess it would be good to clarify why we are telling to use JS symbols as a token

micalevisk (changes requested):
minor improvements

content/fundamentals/dependency-injection.md
Krishan27: Good point. I’ll clarify that symbols provide a unique runtime identity, which helps prevent accidental token collisions compared with string tokens, especially in libraries or larger applications. I’ll also mention that the same exported symbol instance must be reused for registration and injection.

micalevisk (approved):

PR #3500, opened 2026-08-26
docs(recipes): add drizzle orm recipe

alifaroo-q:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Docs

Issue Number: N/A

The docs have recipes for MikroORM, TypeORM, Sequelize, Mongoose and Prisma, but none for Drizzle ORM. There is no guidance on wiring Drizzle into Nest: owning the connection pool, closing it on shutdown, exposing the database through DI, or composing repository methods into a transaction.

## What is the new behavior?

Adds `content/recipes/drizzle.md`, a Drizzle ORM recipe in two parts.

The first part goes from an empty project to a working REST API:

- installing `drizzle-orm`, `pg` and `drizzle-kit`
- defining a schema in plain TypeScript, with row types inferred via `InferSelectModel` / `InferInsertModel`
- configuring `drizzle.config.ts` and generating and applying migrations
- a `PoolProvider` that owns the `pg` pool and closes it in `onModuleDestroy`
- a `DrizzleService` exposing the database, plus named `DrizzleDatabase` / `DrizzleTransaction` / `DrizzleExecutor` types
- a global `DalModule`, a repository, a controller, and a standalone migration script

The second part covers patterns that start to matter as a schema grows:

- shared column blocks (`timestamps`, soft delete)
- branded id types, so a `UserId` cannot be passed where a `PostId` is expected
- repository methods taking an optional `executor`, so each one works standalone or joins a caller's transaction
- a `UnitOfWorkService` for operations spanning several repositories
- optional filters built as `SQL[]` and combined with `and(...)`
- keyset pagination
- `innerJoin` / `leftJoin` nullability
- parameterised raw SQL, and why `sql.raw()` is unsafe
- translating unique-constraint violations into HTTP errors
- testing, using a real database rather than a mocked query builder

Registers the page in `recipes.routes.ts` and the sidebar, and links it from `techniques/sql.md` next to the other ORM recipes.

### Versions

Targets `drizzle-orm@0.45.x` with `drizzle-kit@0.31.x`, the current stable releases. A closing section lists what changes in 1.0 (currently `1.0.0-rc.4`), so the recipe does not go stale silently.

### Verification

- Every TypeScript snippet was compiled against `drizzle-orm@0.45.2` under `strict`, in a scratch project mirroring the recipe's directory layout. No `any` or `@ts-ignore` was added to make anything pass.
- Runtime claims were checked against the Drizzle source rather than assumed. The note that the migrator takes no lock, and so must not run from several replicas at once, comes from reading `migrate()` in `pg-core/dialect.ts`.
- The schema uses `uuidv7()` (Postgres 18) rather than `defaultRandom()`, because the keyset pagination section depends on primary keys sorting in insertion order. The recipe notes the `defaultRandom()` fallback for older Postgres.
- `npm run docs` and `ng build` both pass, all `@@filename` blocks and in-page anchors render, and no unescaped braces reach the Angular templates.

- [ ] Yes
- [x] No

## Other information

`@nestjs/drizzle` does not exist, so unlike the MikroORM recipe there is no third-party module to lean on. The recipe therefore shows the DI wiring explicitly, which is also why it spends time on pool lifecycle and transaction composition.

Happy to trim the patterns half, or split it into a separate page, if that reads as too long for a single recipe.


alifarooq-zk:
Hey @kamilmysliwiec can you review this PR, do you need me to change anything from it.

micalevisk (approved):

Issue #549, opened 2019-07-27
Remove `src/app/pages/*` components


BrunnerLivio:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[x] Feature request
[ ] Documentation issue or request (new chapter/page)
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

At the moment there are tons of Angular components dedicated for each markdown page. The components do not serve any further purpose, they just exist to enable routing and render the HTML template. There are a few minor exceptions:

Not quite sure why these components still contain properties:

- MixinComponents
- SchemaStitchingComponent
- CustomTransportComponent

That would definitely need to be refactored with my proposed changes:

- WhoUsesComponent

## Expected behavior

**This approach is highly inspired by how `angular.io` serves its documentation.**

### Markdown Compilation
All the `src/app/pages/*` components should be removed. The markdown files should be compiled into a static folder e.g. `src/generated/docs`, taking its initial folder structure into account. Therefore e.g.  `content/cli/overview.md` will be compiled to `src/generated/docs/overview.html`.

### <a id="route-handling"></a> Route Handling

A custom `RouterHandlerServce` could then load the HTML files dynamically from the static folder e.g. `http://doc.nestjs.com/cli/overview` will request `http://doc.nestjs.com/generated/docs/cli/overview.html` and render it in a `BaseComponent`.

### Angular Components inside compiled Markdown Files

The generated markdown files still depend on Angular Components/Directives such as `appAnchor` `app-tabs` etc. We still could make use of these components by converting them to native custom elements. angular.io does the same thing [here](https://github.com/angular/angular/tree/master/aio/src/app/custom-elements).

## What is the motivation / use case for changing the behavior?
Why jump over these hoops you may ask? Here are some reasons why this change would be useful:

- **Easier to contribute**
  It is confusing as a user if you need to add a new documentation page. You would have to add a markdown page in `content/` and then in `src/app/pages` and also attach it to the routing.

- **Easier to make changes**
  SEO or general improvements could be changed by editing the one component, which loads in all the compiled markdown pages. (`BaseComponent`)

- **Faster**
  Less to bundle and less to load on the initial request. Lazy loaded by design

- **Useful for the future API docs**
[...]

PR #3514, opened 2026-08-28
feat(docs): add prev/next navigation at the bottom of articles

### LINKED ISSUE #192: Add prev/next arrows at the bottom of articles
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[x] Feature request
[ ] Documentation issue or request (new chapter/page)
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

When I finish reading an article from a group, it's a dead-end. 

## Expected behavior

I can just click "next" to keep reading related articles.

## Minimal reproduction of the problem with instructions

Open any docs page.

It encourages me to continue reading and exploring. Currently I get the feeling "ok and?" as I finish reading. This is especially prominent in [short pages](https://docs.nestjs.com/cli/overview). Would love seeing a link "[Next: **Usage**](https://docs.nestjs.com/cli/usages)" at the bottom.

For example, see [Redux docs](https://redux.js.org/advanced).

micalevisk:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] Docs

Issue Number: closes #192

Finishing an article is a dead end. There is nothing at the bottom of a page that carries the reader onward, which is most noticeable on short pages such as `/cli/overview`. The only way to continue is to go back to the sidebar and find the next entry by hand.

An earlier comment on the issue suggested this was blocked on #549 (moving the route configuration into the Dgeni compiler). That turns out not to be necessary: the sidebar tree is already an ordered structure, so it can be shared as-is and used as the reading order.

## What is the new behavior?

A prev/next pager at the bottom of every documentation article, in the reading order the sidebar already implies.

**Where the order comes from.** The nav tree was moved out of `menu.component.ts` into a shared `nav-items.ts`, now consumed by both the sidebar and the pager, so the two cannot drift apart. The extraction is byte-for-byte faithful; no title, path, ordering, nesting or flag changed. `nav-sequence.ts` flattens that tree into a linear reading order, so the last page of a chapter leads into the first page of the next.

**Chapter context.** Each link shows the chapter it leads into, not just the page title. Several pages share a title — five are called "Overview", and `/pipes` and `/websockets/pipes` sit in identically named neighbourhoods and would otherwise render an identical pager — so the chapter name is what makes a link unambiguous, both visually and in its accessible name.

**Entries that are not "the next thing to read" are excluded.** External-only links (Examples, API Reference, Official courses, Jobs board) are skipped, and pages outside the sidebar (`/enterprise`, `/consulting`) get no pager at all rather than a guess. Lookups use the URL after redirects, so old aliases and `/v11/...` version prefixes inherit the correct pager for free.

**Accessibility.** Because the pager lives outside the router outlet, the link that was just activated survives the navigation and would otherwise keep focus at the bottom of the newly loaded page. Focus is now moved into the article on navigation, and the page is announced politely. This PR also adds the `<main>` landmark and a skip link the site was missing, and gives the sidebar `<nav>` an accessible name so the two navigation landmarks can be told apart.

**Tests.** `nav-routes.spec.ts` resolves the full route table, including lazy chapters, and asserts that every page in the reading order points at a real, non-redirecting route. That check caught a stale `/recipes/swagger` entry: a redirect alias with no content behind it, harmless as a sidebar shortcut but which sent the pager out of Recipes into OpenAPI with a CLI page offered as the way back. The entry is removed here, and the spec keeps that class of bug from returning as pages are added.

- [ ] Yes
- [x] No

## Other information

The `/recipes/swagger` sidebar entry is removed rather than repointed, since the OpenAPI chapter already covers the same material. If you would rather keep it as a shortcut, pointing it at `/openapi/introduction` works too, but it then needs excluding from the reading order so the pager does not traverse it.

Two pre-existing issues were noticed while working on this and deliberately left alone, as they are unrelated to this feature and deserve their own fixes:

- `/enterprise` is unreachable: `{ path: 'enterprise', redirectTo: 'consulting' }` is declared before the route that renders `EnterpriseComponent`, so it shadows it.
- `app.config.ts`'s `provideRouter(routes)` is dead code; `main.ts` bootstraps with `RoutingModule`.

Sourcing the navigation from Dgeni (per #549) remains a reasonable follow-up. Worth noting for whoever picks it up: the content pipeline already registers a reader for `content/navigation.json`, a file that does not exist in the repo.

---

thanks to Opus 5


kamilmysliwiec:
lgtm

PR #3527, opened 2026-09-08
docs(graphql): add Batch loading chapter for @BatchResolveField()

Manuel-Antunes:
Documents the `@BatchResolveField()` decorator proposed in **nestjs/graphql#4123**.

> This PR only makes sense alongside that one — it documents an API that does not exist in a released version yet, so it should not be merged before (or without) it.

## Motivation

Spring for GraphQL ships [`@BatchMapping`](https://docs.spring.io/spring-graphql/reference/controllers.html), which hands a controller method every parent of the current execution layer at once and wires the batching for the developer. The companion PR brings the same capability to Nest, so that solving the n+1 problem no longer requires hand-rolling DataLoaders — instantiating them per request, threading them through the GraphQL context, and re-implementing the positional key/value mapping in every field.

That capability needs a chapter of its own: the n+1 problem is one of the first things people hit when their graph grows, and today the docs have nothing to point them at.

## What this adds

A new **Batch loading** chapter at `/graphql/batch-loading`, placed after "Field middleware" in the GraphQL section:

1. **The n+1 problem** — builds on the `AuthorsResolver` from the Resolvers chapter and shows how 100 authors turn into 101 queries
2. **DataLoader** — what it is, how batching per event-loop tick and memoization work, and why a loader must be per request
3. **Installation** — `dataloader` as an optional peer dependency
4. **Getting started** — the same resolver converted, and the three things that differ from `@ResolveField()`
5. **Mapping the results back to the parents** — `Map` by parent reference, `Map` + `keyBy`, positional array
6. **Parents with no value** — why a missing entry becomes `null`, and how to seed the map
7. **Field arguments** — aliases with different arguments batch separately
8. **Configuring the loader** — options table, plus the `keyBy` vs `cacheKeyFn` distinction
9. **Error handling** — whole-batch failure vs. returning an `Error` for a single parent
10. **Enhancers and field middleware** — middleware runs once per field, enhancers once per batch
11. **Schema first** — no type function needed
12. **Things to keep in mind** — batches follow the event loop, what `@Info()` refers to, and when a plain `@ResolveField()` is still the better choice

## Files

- `content/graphql/batch-loading.md` — the chapter
- `src/app/homepage/pages/graphql/batch-loading/batch-loading.component.ts` — page component
- `src/app/homepage/pages/graphql/graphql.routes.ts` — route
- `src/app/shared/nav/nav-items.ts` — sidebar entry

## Verification

- `npm run docs-only` generates the page cleanly (137 docs). The only link the dangling-link report flags is `/graphql/other-features#execute-enhancers-at-the-field-resolver-level`, which is the identical link `field-middleware.md` already uses.
- The generated template's dependencies (`app-copy-button`, `app-tabs`, `appAnchor`, `routerLink`, `extension` pipe) match the component's `imports` — a strict subset of what `resolvers-map.component.ts` declares.
- Every behavioral claim in the chapter is backed by a test in the companion PR, including the two least obvious ones: field middleware running once per field around the batch, and enhancers running once per batch with the array of parents as the execution-context root.

### DIFF
--- a/content/graphql/batch-loading.md
+++ b/content/graphql/batch-loading.md
@@ -1,0 +1,295 @@
+### Batch loading
+
+Field resolvers let you compose a graph out of small, focused functions. They also make it easy to accidentally hammer your data source: GraphQL calls a field resolver **once for every parent object**, so a query that returns 100 authors calls the `posts` field resolver 100 times.
+
+This is known as the **n+1 problem**. The `@BatchResolveField()` decorator solves it by handing your method **every parent at once**, so you can load the whole layer with a single round-trip.
+
+#### The n+1 problem
+
+Take the resolver from the [resolvers](/graphql/resolvers) chapter, adjusted to return a list of authors:
+
+```typescript
+@@filename(authors/authors.resolver)
+@Resolver(() => Author)
+export class AuthorsResolver {
+  constructor(
+    private authorsService: AuthorsService,
+    private postsService: PostsService,
+  ) {}
+
+  @Query(() => [Author])
+  authors() {
+    return this.authorsService.findAll();
+  }
+
+  @ResolveField(() => [Post])
+  posts(@Parent() author: Author) {
+    return this.postsService.findByAuthorId(author.id);
+  }
+}
+```
+
+Now run this query:
+
+```graphql
+{
+  authors {
+    firstName
+    posts {
+      title
+    }
+  }
+}
+```
+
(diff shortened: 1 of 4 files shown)


micalevisk (approved):

kamilmysliwiec:
This PR was merged by accident during a batch merge that was meant for dependency updates only, so it's being reverted in #3543 (along with the companion nestjs/graphql#4123). Sorry for the noise. We'll revisit it separately with a proper review.

